PR #2, opened 2020-04-20
cmake for visual studio

tuanhtrng:
These boys can build uox3 in one go on windows with vs. They will not work on Linux machine but not that hard to make it work on Linux since I have a working cmake set on OpenBSD.


tuanhtrng:
Does not work!

PR #3, opened 2020-04-25
using template

tuanhtrng:
It was compiled but not tested.


Xoduz:
Closed on request of tuanhtrng

PR #16, opened 2021-06-08
Thirst System

DragonSlayer62:
This will take stamina if the player gets thirsty this can be toggled on or off.


Xoduz (approved):
Good stuff! A couple of things missing to make it fully functional - there was no call to the DoThirst() function, for instance, and the thirst system was using the hunger timer instead of the new thirst-specific one that was added. I sorted those small issues, and added some extra stuff - like thirst-related JS properties for characters, made drinking drinks reduce thirstiness - as well as a couple of style tweaks here and there. Now it's good to go!

PR #18, opened 2021-06-12
Merge of Dabble

punt1959:
So to test out the whole concept. All this does is NULL to nullptr where appropriate. And replaces Config.h with ConfigOS.h. UOX_PLATFORM went to PLATFORM (and then PLAFORM_WIN32 went to WINDOWS, etc).

So simple changes, just testing to get an understanding of this pull


Xoduz (approved):
Good stuff. Merging as soon as checks clear :)

PR #30, opened 2021-09-20
Complete Cooking Overhaul

DragonSlayer62:
New Folder foods
Moved drinks.dfn to foods
moved all food items to foods.
split food2 into different cats
Updated all cooking js files
Added heatsources.js
fixed flour mill to have motion when used also to stop and be double clicked to get your flour bag
Fixed flour bag to have 20 uses of flour
Fixed water pitcher to set the 20 uses of flour on the bag if water is pour from it. also not to delete but use some content.
Fixed dough now can make all pies and sweet dough.
Added cooking.js that you can cook all types of meat.
Added baking.js for dough.
All is On Par with Pub 15. They did not add a Gump for cooking until pub 16.


DragonSlayer62:
this was added to wrong dev i am closing will re pr later

PR #44, opened 2022-01-26
Yetagain

punt1959:
Added a VS2022 in top level (not ideal, as projects (zlib/jsengine), should be in their own project folders.  Removed two defines in ConfigOS.h given their build specifics, (which would need to be added the VS2017/2019 UOX3 project c++ proprocessor options)


Xoduz:
Closed on request.

PR #128, opened 2022-10-24
FreeBSD build support

ldilley:
### Proposed Changes:
- Fixed `cPlayerAction.cpp` case in `source/CMakeLists.txt` (was `cplayeraction.cpp`).
- Added FreeBSD build support for SpiderMonkey, zlib, and UOX3.
- Updated `README.md` with FreeBSD build instructions.
- Fixed the following compile error due to `NULL` being defined as `nullptr` (non-int type) and not `0` (zero):
 ```sh
 SEFunctions.cpp:3500:10: error: assigning to 'jsval' (aka 'long') from incompatible type 'nullptr_t'
        *rval = NULL;
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,14 @@ Join the [UOX3 Discord](https://discord.gg/uBAXxhF) for support and/or a quick c
 
 First step, open a new terminal and enter the commands below:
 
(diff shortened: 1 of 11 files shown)


ldilley:
This fork was also built successfully using GCC 10.2.1 under Debian 11.4. It was originally built using Clang 13.0.0 under FreeBSD 13.1-RELEASE. Lastly, the `uox3` binary ran as expected.

Xoduz (approved):

PR #136, opened 2022-10-30
Taxidermy Kit

DragonSlayer62:
So This was not to hard to create. and I think uox3 users will love to have it.


Xoduz (changes requested):
Looks good! But a couple of points:
- How do players get hold of taxidermykits?
- You can't really rely on corpse names to identify different creatures that can have trophies created - those names can change based on language used by server. Use the amount tag (for creature ID) combined with .sectionID. After my next commit, sectionID of NPC will be stored on the NPC's corpse when they die.
- Item definitions for the actual trophies should inherit base_item via the GET tag, and should maybe also include something like a movement=3 tag to ensure they are non-movable.

Xoduz (approved):

PR #137, opened 2022-10-30
House Decay Added.

DragonSlayer62:


Xoduz (changes requested):
A few points to address:
- Follow camelCase naming convention for variables (DecayEnabled)
- Follow style guidelines with regards to spaces for if/for statements:
`if( blah || ( bleh ))`
`for( something )`
instead of
`if ( blah || ( bleh ) )`
`for ( something )`
- Remove reliance on storing character who enters on multi's .tempObj. Not needed. 
- For the Console message in HouseDecay, don't need to specify language. 
- When placing items on the ground, can place them at the ground level of the house's ban location - you don't know that the last player who entered the house is even in the same world any more at the time the house decays, they could be long gone!
- Don't kill all timers on the multi indiscriminatingly and restart them - you don't know what other scripts are running timers! Use .SetJSTimer() to update the remaining time instead
- How does a house owner know what the decay state of their house is? Should some form of decay state be visible to house owner?
- Ideally something like opening the doors of the house, or opening the containers, should also refresh the house - owner might just log on while in the house, never leave, do a bunch of stuff and log back out. If you turn the "refresh timer" portion into a separate function, you can call that function using TriggerEvent from a script like door.js if you detect players are in a house they own/co-own/are friended to.
- Is there a reason to keep the house keys for previous house around after it has decayed?

PR #150, opened 2022-11-29
Various fixes, improvements, and documentation updates

ldilley:
## Proposed Changes
- Comment line in `jsconfig.mk` that attempts to `cat` a non-existent `Version` file. This causes an error otherwise during the SpiderMonkey build.
- Initialize `totalPlayerGold` in `vendor.cpp`.
- Flag unused variable in `gumps.cpp`.
- Update `README.md` to include references to FreeBSD and also alter the command (`libtool` instead of `ar`) used to generate the JavaScript library on macOS.
- Improve `automake.sh`.

### DIFF
--- a/automake.sh
+++ b/automake.sh
@@ -66,6 +66,7 @@ fi
 ev=$?
 if [ $ev -ne 0 ]; then
   echo "Unable to build UOX3! Exiting..."
+  cd ..
   exit $ev
 else
   if [ -f ./uox3 ]; then
@@ -78,4 +79,3 @@ else
   fi
 fi
 cd ..
-exit $ev
--- a/automake.sh
+++ b/automake.sh
@@ -1,5 +1,6 @@
-#!/usr/bin/env bash
-echo "Building spidermonkey"
+#!/usr/bin/env sh
+
+echo "Building SpiderMonkey..."
 cd spidermonkey
 if [ "$(uname)" = "FreeBSD" ]
 then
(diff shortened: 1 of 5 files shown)


source/gumps.cpp
```diff
@@ -1907,7 +1907,7 @@ bool CPIGumpMenuSelect::Handle( void )
 	Console << "Type is " << gumpId << " button is " <<  buttonId << myendl;
 #endif
 
-	CItem *j = nullptr;
+	[[maybe_unused]] CItem *j = nullptr;
```
strattond, line 1910: Lose it if it's not used, and it looks not used to me

source/vendor.cpp
```diff
@@ -120,7 +120,7 @@ bool ApplyItemSection( CItem *applyTo, CScriptSection *toApply, std::string sect
 bool CPIBuyItem::Handle( void )
 {
 	UI16 i;
-	UI32 totalPlayerGold, totalGoldCost = 0;
+	UI32 totalPlayerGold = 0, totalGoldCost = 0;
```
strattond, line 123: 👍

spidermonkey/jsconfig.mk
```diff
@@ -82,7 +82,7 @@ ifeq ($(OS_ARCH), WINNT)
   NSPR_SHARED    = nspr20/$(NSPR_VERSION)/$(NSPR_OBJDIR)
 endif
 NSPR_VERSIONFILE = $(NSPR_LOCAL)/Version
-NSPR_CURVERSION := $(shell cat $(NSPR_VERSIONFILE))
+#NSPR_CURVERSION := $(shell cat $(NSPR_VERSIONFILE))
```
strattond, line 85: Any particular reason?

automake.sh
```diff
@@ -8,54 +9,73 @@ else
   make -f Makefile.ref DEFINES=-DHAVE_VA_LIST_AS_ARRAY CC=gcc
 fi
 
-$ev=$?
+ev=$?
```
strattond, line 12: Thanks, I had this to fix up

automake.sh
```diff
@@ -8,54 +9,73 @@ else
...
 else
   make
 fi
+
 ev=$?
-if [ -f ./uox3 ]; then
-  cp uox3 ..
-  echo "Done! You should now find the compiled uox3 binary in the root UOX3 project directory. Copy this binary to a separate directory dedicated to running your UOX3 shard, along with the contents of the UOX3/data directory, to avoid potential git conflicts and accidental overwriting of data when pulling UOX3 updates in the future."
+if [ $ev -ne 0 ]; then
+  echo "Unable to build UOX3! Exiting..."
```
strattond, line 68: If you don't `cd ..` here, you'll end up in a different folder than you start

automake.sh
```diff
@@ -8,54 +9,73 @@ else
...
+    cp uox3 ..
+    echo "Done! You should now find the compiled uox3 binary in the root UOX3 project directory. Copy this binary"
+    echo "to a separate directory dedicated to running your UOX3 shard, along with the contents of the UOX3/data directory,"
+    echo "to avoid potential git conflicts and accidental overwriting of data when pulling UOX3 updates in the future."
+  else
+    echo "uox3 program not found! Please review the build status."
+  fi
 fi
 cd ..
 exit $ev
```
strattond, line 81: You can lose the `exit $ev` here because you're terminating on ev above.

automake.sh
```diff
@@ -8,54 +9,73 @@ else
...
+  echo "Unable to build UOX3! Exiting..."
+  exit $ev
 else
-  echo "uox3 output not created!  Please review compile status"
+  if [ -f ./uox3 ]; then
+    cp uox3 ..
+    echo "Done! You should now find the compiled uox3 binary in the root UOX3 project directory. Copy this binary"
+    echo "to a separate directory dedicated to running your UOX3 shard, along with the contents of the UOX3/data directory,"
+    echo "to avoid potential git conflicts and accidental overwriting of data when pulling UOX3 updates in the future."
+  else
```
strattond, line 76: How do you envisage this else ever triggering?  How could make pass, but not result in a valid output?

strattond (changes requested):
Overall great improvements, just couple of minor comments/questions and a suggestion to remove rather than mark unused

source/gumps.cpp
ldilley: It looks like @punt added this 2 years ago. I've just removed it.

spidermonkey/jsconfig.mk
ldilley: Yes; as noted in the proposed changes, the NSPR `Version` file never gets created. This line attempts to `cat` a non-existent file twice which causes 2 unnecessary errors during the build process of SpiderMonkey. This line is not required and there is no harm in commenting it.

automake.sh
ldilley: Added in 71cc771.

automake.sh
ldilley: Removed in 71cc771.

automake.sh
ldilley: Are you suggesting removal of the file check prior to copying too? I say it is better safe than sorry and to not make any assumptions. It's like having a default safety net to handle the possibility. Perhaps some nameless security tool flags `uox3` as malicious (because the program is unknown and listens for inbound connections on a socket) and removes the binary immedlately following its generation. If the file can exist, it also seems logical that the opposite case could conceivably occur. Lastly, it is merely two additional lines in a shell script -- I do not feel there is any harm in keeping the `else` and `echo` lines intact.

Xoduz: Resolving this for now, can revisit later if needed.

Xoduz (approved):

PR #154, opened 2022-12-03
Fix size_t -> int issues with casts

ldilley:
## Proposed Changes
- Fix Clang 14.x `size_t` -> `int` warnings on macOS reported by @punt1959.


strattond (approved):
This will do it, but it feels like we're just removing warnings for the sake of warnings.  Beyond some noise, this isn't going to do a lot.

Casting it to a signed value will half our potential size (though never going to go near it)

ldilley:
> This will do it, but it feels like we're just removing warnings for the sake of warnings. Beyond some noise, this isn't going to do a lot.

Fixing warnings was the entire point of this endeavor. That objective is being met (by PRs #138 and #150 as well). This PR is not intended to change functionality in any significant manner. This particular PR was a mere 9 lines of small changes -- It isn't meant to do a lot. Although some folks may not feel the same, I believe that this is a worthwhile effort.

PR #158, opened 2022-12-11
Solen Ant Hole Add-On

DragonSlayer62:
- [ADD] Added Solen Ants black and red
- [ADD] Added Solen Spawns for Felucca and Trammel
- [ADD] Added Solen hive teleporters
- [ADD] Added Solen Hive Dungeon Trammel Region
- [FIX] Fixed an issue with admin welcome gump where checkbox ID for decoration addons in Trammel were off by 1, resulting in wrong decorations being loaded
- [UPD] Resized admin welcome gump to accommodate additional decoration addons in Trammel


data/dfndata/house/house.dfn
```diff
@@ -3581,6 +3638,260 @@ Y=0
 Z=0
 }
 
+// dirt 1
+[HOUSE ITEM 600]
```
strattond, line 3642: Is there a particular reason why the dirt ones are ever so slightly different in position?

data/dfndata/items/houseaddons/solenanthole.dfn
```diff
@@ -0,0 +1,233 @@
+// solen ant hole bits and pieces
+
+[dirt1]
```
strattond: Worth a [basedirt] that has a name, weigh, movable and decay commonality?

data/dfndata/items/houseaddons/solenanthole.dfn
```diff
@@ -0,0 +1,233 @@
...
+NAME=a hole
+ID=0x1776
+COLOUR=0x1
+weight=100
+movable=2
+decay=0
+SCRIPT=2752
+}
+
+[shentrancepeice1]
```
strattond: Ditto for a [base_shentrancepiece] - also, you've spelt piece wrong :)

data/dfndata/npc/hymenoptera_base.dfn
```diff
@@ -0,0 +1,373 @@
...
+BACKPACK
+SKIN=0x453
+RACE=31
+NPCWANDER=0
+ORIGIN=lbr
+}
+
+[ambitious_black_solen_queen]
+{
+GETUO=ambitious_black_solen_queen_lbr
```
strattond, line 26: So the implication is that this LBR+ based on the name.

Should there be a GETUO, GETT2A, GETUOR or GETTD if they're prior?  What happens if they're missing?

data/dfndata/npc/npc_aos/hymenoptera_aos.dfn
```diff
@@ -0,0 +1,291 @@
...
+RESISTLIGHTNING=25 30
+TOPROV=796
+TOPEACE=796 5
+}
+
+[black_solen_warrior_aos]
+{
+GET=base_black_solen_warrior
+GOLD=250 350
+//Bracelet of Binding needs created
```
strattond: So ... where is it?  Either of these PACKITEMs don't seem a match?

data/js/server/teleporter/teleporter_solen_hive.js
```diff
@@ -0,0 +1,332 @@
...
+	"5758,1795": [5725, 1894],
+
+	//B leads back level 1
+	"5727,1894": [5757, 1798],
+	"5728,1896": [5757, 1798],
+	"5728,1895": [5757, 1798],
+	"5728,1894": [5757, 1798],
+
+	//C leads to center area
+	"5657,1898": [5750, 1894],
```
strattond, line 59: So this is "C leads to center area", but 2 above with the same target coordinates is "B leads to level 2"

Are "center area" and "level 2" the same?

data/js/server/teleporter/teleporter_solen_hive.js
```diff
@@ -0,0 +1,332 @@
...
+*	We also have the Main Solen hive Entrances.
+*	Description how it works
+*	1. You double click it will take you down to the solen hive depending on what location you are using.
+*	2. Double click in the middle of the black inside the solen hive it will return you to the surface.
+*
+*	World Decorations that come with this file will place the Main Entrances and Exits around the World for Felucca Map only.
+*/
+
+// Entrance coordinates
+var entranceMap = {
```
strattond, line 20: Wondering if it's worth pulling targets out as meaningfully named constants and using those constants in the key value?
e.g. with the Solen Hive teleporters, there's lots that go to the same spot.

```
var SOLEN_LEVEL_2 = [5725, 1894];
...
"5757,1794": SOLEN_LEVEL_2,
```

strattond (changes requested):
More questions than critiques

data/dfndata/spawn/trammel/spawn_trammel_dungeon_solen_hives.dfn
```diff
@@ -0,0 +1,62 @@
...
+[REGIONSPAWN 14300]
+{
+NAME=Red Solen Hive
+NPCLIST=red_solen
+MAXNPCS=100
+X1=5645
+Y1=1781
+X2=5929
+Y2=2034
+WORLD=0
```
Xoduz: World number is incorrect on these spawn regions. By the name of the file they look intended for trammel?

data/dfndata/npc/hymenoptera_base.dfn
Xoduz: Originally for LBR+, but no reason why people shouldn't be able to enable it for earlier eras if desired. But the way the system is currently setup, it does require that tag for each era, or a broken NPC will spawn instead.

Xoduz (changes requested):

data/dfndata/house/house.dfn
DragonSlayer62: Yes they make a tunnel

image

image

data/dfndata/items/houseaddons/solenanthole.dfn
DragonSlayer62: image
image


this creates the tunnels two differant versions

data/dfndata/items/houseaddons/solenanthole.dfn
DragonSlayer62: will fix thanks

data/dfndata/npc/npc_aos/hymenoptera_aos.dfn
DragonSlayer62: this not scripted yet comment for when it does get added

data/js/server/teleporter/teleporter_solen_hive.js
DragonSlayer62: https://[uo.stratics.com/content/atlas/solenhive.php](https://uo.stratics.com/content/atlas/solenhive.php) so the dungeon is just really one level but broken into 4 sections that have different teleport points.

data/js/server/teleporter/teleporter_solen_hive.js
DragonSlayer62: well, they go to the same area just different parts of that area

data/dfndata/spawn/trammel/spawn_trammel_dungeon_solen_hives.dfn
DragonSlayer62: Fixed

data/dfndata/items/houseaddons/solenanthole.dfn
strattond: Yeah, my point was more that for dirt1 -> dirt9, the NAME, weight, movable and decay values are all the same.  It's only ID and possibly COLOUR that changes.

You could move the NAME, weight, movable and decay into a common basedirt and call it a day?

In fact, looking at solenentrancehole and it's variants, the weight, movable and decay values match there too.

So you could have a "base_immovable" that is the weight, movable and decay and GET those where appropriate.  For example:

```
[base_immovable]
{
weight=100
movable=2
decay=0
}

[ahole]
{
NAME=a hole
ID=0x122A
COLOUR=0x1
GET=base_immovable
SCRIPT=2752
]
```

DragonSlayer62: I see what you mean I fixed it.

strattond (approved):

data/dfndata/items/houseaddons/solenanthole.dfn
Xoduz, line 1: If these items are intended specifically for the solen ant holes, rename them. Either add the dirt items as generic items in a more generic file, or rename them to be specifically for the solen ant holes. Example: [base_anthole_dirt], [anthole_dirt1], [anthole_dirt2], {anthole], etc.

data/dfndata/npc/npc_aos/hymenoptera_aos.dfn
Xoduz: I think you can comment out loot like this until the item functionality as been scripted. Don't add "junk items" that do nothing, can't be sold for money, just take up space, etc.

data/dfndata/spawn/felucca/spawn_felucca_dungeon_solen_hives.dfn
```diff
@@ -0,0 +1,62 @@
+// spawn_felucca_dungeon_solen_hives.dfn
+// [4300-4349]
+
+[REGIONSPAWN 4300]
```
Xoduz, line 4: No. This is lazy design, and risks setting up situations where too many NPCs spawn in a single place, since it's all random across the entire dungeon. Split this up into additional spawn regions, 6 of them, each with MAXNPC=16 or some such. Same for Trammel spawns. Something like this:

image

data/js/jsdata/worldtemplates/felucca_solenhive.jsdata
Xoduz, line 1: This is good stuff, but... it's not hooked up to anything, so these items will never spawn unless someone does 'decorate load felucca_solenhive

Ideally, this would be part of the admin welcome gump, as an option for additional decorations for Felucca/Trammel:

image


Also, the template file includes several "deed" items (id 5360), which spawn very visible in the world? I suggest removing these from the file.

data/js/server/teleporter/teleporter_solen_hive.js
```diff
@@ -0,0 +1,332 @@
...
+			pUser.SysMessage( "I can't reach that." );
+			return false;
+		}
+		
+		// Use the tile coords as the lookup key
+		var key = iUsed.x + "," + iUsed.y;
+
+		// Try entrance map first
+		if( teleportFromMap( pUser, key, entranceMap ))
+			return true;
```
Xoduz: "return true" here causes a system message that says "You can't think of a way to use that item." because it tries to pass the double-click event to code, and finds nothing actionable. Change to return false (as well as for the next one below.)

data/js/server/teleporter/teleporter_solen_hive.js
```diff
@@ -0,0 +1,332 @@
+/*
+*	Solen Hives and Ants was introduced in LBR during a event called When Ants Attack.
+*
+*	We have the Randomly spawning black holes that can be put out with event tags for events.
```
Xoduz, line 4: Are these ant holes randomly spawned? Should there be spawn logic that makes this happen? Are they currently just fixed ant holes in every single possible location?

Xoduz (changes requested):

data/dfndata/items/houseaddons/solenanthole.dfn
DragonSlayer62: Ill rename them because there tied to a script

data/js/server/teleporter/teleporter_solen_hive.js
DragonSlayer62: the ones in decoration file are fixed the random ones is the ones a gm can use to put out for players to use and those will randomly drop you someone in the dungeon.

Xoduz (approved):

PR #160, opened 2022-12-17
Shopkeeper distance fix

Xoduz:
When a player says "vendor buy" or "vendor sell", NPC shopkeepers (and related JS events) will now respond in priority of distance to the player


ldilley:
@Xoduz and I briefly discussed adding `const` to the pair of items below. However, doing so would impact other parts of the code.

```cpp
( CChar *a, CChar *b )
```

ldilley (approved):

PR #161, opened 2022-12-19
Added Map Ruleset

DragonSlayer62:
This can make it so players can not attack steal or cast on other players why in trammel ,malas and  Tokuo or ilsh.

On login will attach the ruleset to all the characters logging in to the game. This can also be used later for the young player status as well.


Xoduz (approved):
Good enough for now. Let's push it to develop, and test it :)

PR #163, opened 2022-12-31
Misc updates

Xoduz:
Extended GetSpawnRegion JS Function to fetch spawn region reference based on a set of coordinates. Supported syntax are now:
-	GetSpawnRegion( spawnRegionID )
-	GetSpawnRegion( x, y, worldNum, instanceID )

Added new DFN tags for Items and NPC:
-	SECTIONID	// Override which [sectionID] is stored for an object
-	ORIGIN		// Store reference to which expansion/era of UO an item/NPC originated in

Added new Item/Character JS property:
-	.origin 	// Get reference to expansion/era of UO an item/NPC originated in

Fixed an issue with IsInBuilding JS Function which could only handled a limited subset of instance ID values Corrected name of tag used to attach scripts to ore types - from SCRIPTID to SCRIPT


ldilley (approved):
This looks good to me. Happy New Year also, @Xoduz.

PR #164, opened 2023-01-09
Docker build (not run)

strattond:
First phase of docker implementation for the build and run process.

Currently building is working, and an image created.  However, no volume or data locations exposed as yet (TBD).


ldilley (approved):
Nice job simplifying the multi-stage build steps (combining the SpiderMonkey and zlib build phases with uox3) and decoupling the heredoc for `run.sh` per our Discord conversation. This will allow the `Dockerfile` to be compatible with older versions of Docker without BuildKit (`buildx`) support along with Podman (until containers/buildah#3474 can be resolved).

PR #176, opened 2023-07-01
UOX3 0.99.6-RC6

Xoduz:
- [x] Removed arbitrary and hidden restrictions on refreshing list of online players shown with 'wholist command
- [x] Fixed an issue where player ghosts that get teleported would not see themselves get updated to new location
- [x] Fixed a server crash from clients disconnecting while running server in debug mode
- [x] Improved how container updates are sent to players; now only sends to clients that have actually opened the container, and who are still within range
- [x] Added more information about UO data files being loaded during server startup
- [x] Enhanced the GM 'add menu to allow clicking directly on images of items/npcs to add them, and adjusted size of menus to accommodate this
- [x] Fixed an issue with fleeing NPCs that could cause them to attempt to flee across time and space, regardless of distance to target/attacker in combat
- [x] Added NPC AI that runs away from players if they get too close (even outside combat). Applied to hind, rabbits, squirrels, ferrets and various birds by default:
  -	AI_ANIMAL_SCARED (aitype 12)
- [x] Fixed issues with NPCs fleeing forever, or getting stuck in flee/don't flee loop, by introducing max fleeing distance (50 tiles) and cooldown (30 sec) on fleeing
- [x] Updated onNameRequest JS Event to include third parameter with origin of name request, so script responses can be tailored appropriately:
  - onNameRequest( myObj, requestedBy, requestSource )
    - Potential values for requestSource:
      - 0 - Speech/System Messages
      - 1 - Guild Menus
      - 2 - Stat Window (self)
      - 3 - Stat Window (other)
      - 4 - Tooltip
      - 5 - Paperdoll Journal
      - 6 - Paperdoll
      - 7 - Single Click / All-Names
      - 8 - System
      - 9 - Secure Trade Window
- [x] Added Adaptive Performance System (APS) that dynamically adjust how often NPC AI/movement is checked based on overall shard performance. If performance drops below defined threshold, UOX3 gradually slows down checks for NPC AI/movement to prioritize player movement/speech/command responsiveness. If performance climbs back up above threshold, slowdowns are gradually removed. The following UOX.INI options have been added under [system] category to support this system:
  - APSPERFTHRESHOLD=50 	// Performance threshold (measured in simulation cycles/sec) below which the APS kicks in
  - APSINTERVAL=100 		// How often (in milliseconds) the APS checks performance and makes adjustments (if needed) to balance out shard performance
  - APSDELAYSTEP=50 		// How much the delay timer is modified by (in milliseconds) each time APS makes adjustments
[...]

PR #177, opened 2023-07-08
Fix compile warnings (for 0.99.6-RC6/PR #176)

ldilley:
* Fixed initialization order
* Marked unused variables
* Fixed loop iterator data type
* Removed useless `const` qualifiers
* Removed extraneous whitespace
* Used additional parentheses in comparisons to make intent clearer

### DIFF
--- a/source/ai.cpp
+++ b/source/ai.cpp
@@ -68,7 +68,7 @@ bool IsValidAttackTarget( CChar& mChar, CChar *cTarget )
 				// Young players are not valid targets for NPCs outside of dungeons
 				if( cwmWorldState->ServerData()->YoungPlayerSystem() && mChar.IsNpc() && !cTarget->IsNpc() && IsOnline(( *cTarget )) && cTarget->GetAccount().wFlags.test( AB_FLAGS_YOUNG ) && !cTarget->GetRegion()->IsDungeon() )
 				{
-					if( cTarget->GetSocket() && (cTarget->GetTimer( tCHAR_YOUNGMESSAGE ) <= cwmWorldState->GetUICurrentTime() || cwmWorldState->GetOverflow()) )
+					if( cTarget->GetSocket() && ( cTarget->GetTimer( tCHAR_YOUNGMESSAGE ) <= cwmWorldState->GetUICurrentTime() || cwmWorldState->GetOverflow() ))
 					{
 						cTarget->SetTimer( tCHAR_YOUNGMESSAGE, BuildTimeValue( static_cast<R32>( 120 ))); // Only send them the warning message once couple of minutes
 						cTarget->GetSocket()->SysMessage( 18734 ); // A monster looks at you menacingly but does not attack.  You would be under attack now if not for your status as a new citizen of Britannia.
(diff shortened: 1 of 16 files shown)


source/ai.cpp
```diff
@@ -68,7 +68,7 @@ bool IsValidAttackTarget( CChar& mChar, CChar *cTarget )
 				// Young players are not valid targets for NPCs outside of dungeons
 				if( cwmWorldState->ServerData()->YoungPlayerSystem() && mChar.IsNpc() && !cTarget->IsNpc() && IsOnline(( *cTarget )) && cTarget->GetAccount().wFlags.test( AB_FLAGS_YOUNG ) && !cTarget->GetRegion()->IsDungeon() )
 				{
-					if( cTarget->GetSocket() && cTarget->GetTimer( tCHAR_YOUNGMESSAGE ) <= cwmWorldState->GetUICurrentTime() || cwmWorldState->GetOverflow() )
+					if( (cTarget->GetSocket() && cTarget->GetTimer( tCHAR_YOUNGMESSAGE ) <= cwmWorldState->GetUICurrentTime()) || cwmWorldState->GetOverflow() )
```
Xoduz, line 71: This changes the flow of the logic 🤔 The check should be:
- "if socket is valid, AND if EITHER tCHAR_YOUNGMESSAGE is lower than GetUICurrentTime OR GetOverflow returns true"

but with this change it becomes:
- "if socket is valid AND tCHAR_YOUNGMESSAGE is lower than GetUICurrentTime, OR if GetOverflow returns true"

source/cItem.cpp
```diff
@@ -2683,8 +2683,8 @@ void CItem::SendToSocket( CSocket *mSock, [[maybe_unused]] bool drawGamePlayer )
 			auto iVisible = GetVisible();
 			auto ownerObj = GetOwnerObj();
 			if(( iVisible != VT_VISIBLE && iVisible != VT_TEMPHIDDEN )
-				|| iVisible == VT_TEMPHIDDEN &&
-				( !ValidateObject( ownerObj ) || ( ValidateObject( ownerObj ) && ownerObj->GetSerial() != mChar->GetSerial() )))
+				|| ((iVisible == VT_TEMPHIDDEN &&
```
Xoduz, line 2686: This, as well, changes the intended logic. It's supposed to be:
- if visible state is VT_TEMPHIDDEN AND either ownerObj exists OR ownerObj exists and owner serial differs from mChar's serial

source/ai.cpp
ldilley: Does commit 9d985de now meet the proper criteria?

source/cItem.cpp
ldilley: Does commit cf28051 now contain the correct logic?

Xoduz: Not quite 🤔The idea is that the checks on ownerObj & the serial comparison only runs if iVisible equals VT_TEMPHIDDEN. So, maybe... this?
```
if(( iVisible != VT_VISIBLE && iVisible != VT_TEMPHIDDEN )
	|| ( iVisible == VT_TEMPHIDDEN &&
	( !ValidateObject( ownerObj ) || ( ValidateObject( ownerObj ) && ownerObj->GetSerial() != mChar->GetSerial() ))))```

source/ai.cpp
Xoduz: It does! Thank you :)

source/cItem.cpp
ldilley: Fixed in f4e7e35.

Xoduz (approved):

PR #179, opened 2023-08-14
Feature/js upgrade

strattond:
Initial work to get 1.8.0-RC1 upgrade of the javascript engine.

Best effort to resolve the jsautocfg issue - CMake builds it and runs it prior to building SpiderMonkey.

XCode updated, but I don't know how to achieve the equivalence from the IDE.


ldilley (approved):
I made a small change to `spidermonkey/make/cmake/CMakeLists.txt` to omit the `-MD` flag on *BSD due to the following output during compilation:
```
cc: warning: argument unused during compilation: '-MD' [-Wunused-command-line-argument]
```

I also updated `spidermonkey/README.md` to reflect the version change to SpiderMonkey 1.8.

This PR builds fine on both FreeBSD using Clang and Linux using GCC.

Xoduz (approved):

PR #181, opened 2023-08-23
xcode updates

OldFoolAgain:
updated to enable Xcode function


strattond:
I see nothing obviously weird here outside the code scope and I'm comfortable the xcode stuff would've been thoroughly tested.

Looks good

PR #182, opened 2023-08-23
cmake updates

OldFoolAgain:
added building zlib as optional (macOS always has installed and built for both architectures). Unix may
or may not have it installed.  Defaults to using system on macOS, and using UOX3 version on unix (and windows). One can 
define SYSTEM_ZLIB on their Cmake line to use the system zlib.

quieted warnings on cmake build. Added some disabling to warnings on the Cmake build to not worry the user.


strattond:
Declining this as it was superseded by the workflow PR

PR #183, opened 2023-08-23
Workflow

OldFoolAgain:
Includes updates from the cmake pull request
Consolidates github workflows to a single one, that will build on all three OS (to include macOS)
(Status can be gotten on the individual builds through the GitHub interface)

Modified readme.md , to consolidate cmake build commands for all OS's.  Also eliminated need to create build directory as the cmake instruction (-B) will auto create it.


strattond (approved):
All eminently reasonable changes, happy to accept

PR #185, opened 2023-08-24
Organization for ease of maintenance when updating ide/make files from different OS's

OldFoolAgain:
This contains five things:
1. Memory patch in RC (was in my version, so sorry)
2. Fixed two warnings under VS (cast a size_t to an int (when it was pushed into a vector<int>, and unrefence use of a variable in a catch() statement.
3. Renamed two files to make them match: cBaseobject.cpp -> cBaseObject.cpp , gump.h -> CGump.h (as it defines the CGump class in CGump.cpp).  Associated files that included/used these files modified.
4. This was the main focus of this update, organization.  Right now, if one wants to manually update project files, and CMake files, there is no organization per se (grouping) to easily find files.  So this organized CMake, and the VS2022 project/filter files to group them together. Now when one needs to update (to support cross platform developers), and have to edit those files with a text editor, there is some logical grouping. 
5. Deleted redundant workflow (build.yml)

NOTE: 
1. It was not the intention of changing groupings, if it did, it was by accident.
2. If you change the grouping in windows, PLEASE change with an editor, not in the ide.  The ide does not sort (group) on writes, so it becomes a mess. It is more effort on the windows users, but the flip side it is no more effort then any non windows developer has to do to support Windows devs.  
3. No VS2017 grouping was made, the only modification was to the two name changes to make them consistent (cbaseobject/cgump)


OldFoolAgain:
VS 2017 now modified, only to accommodate cBaseobject.cpp to cBaseObject.cpp and gump.h to Gump.h change

Xoduz (approved):

PR #186, opened 2023-08-25
Overlap - XURI please test

OldFoolAgain:
For lack of a better branch to merge this into
It contains:
js-upgrade, plus the Organization PR 
this is a test of async region save (very initial, probably not the final way I would do it).
Nothing but region saves are async (and frankly, the future waits should be checked later, so depending how this shows, that would be the progression).

It does have the command entry clear bug I posted about (I don't believe that is anything I introduced, not an area I touched, and was there after the last PR before this one).

Anyway, Xuri, if you can test on your large world save, I would appreciate it.


OldFoolAgain:
Right now, it doesn't leave the region save() until all writes are finished. There is no reason for this, and. would obviously change that. So right now, I am not expecting a large (if much of any) improvment.  However, if you are okay that this is functionally equivalent, I can actually do: 1. All of the writes, and 2. Don't wait until the next attempt tisane (as that is when one needs to wait). So essentially, the write has the entire processing cycle to complete.

PR #208, opened 2023-09-28
MakeItem in Document Update

DragonSlayer62:
Updated Makeitem Doc add in Hue color can be added to this.

### DIFF
--- a/docs/jsdocs.html
+++ b/docs/jsdocs.html
@@ -7259,7 +7259,7 @@ if( GetRaceSkillAdjustment( myRace.id, 32 ) > 10 )
 							<div class="spoiler">
 								<div class="settingsDiv">
 									<p><span class="hl">Prototype</span></p>
-									<p><em>MakeItem( socket, player, itemNum );</em></p>
+									<p><em>MakeItem( socket, player, itemNum, hueNum);</em></p>
 								</div>
 								<div class="settingsDiv">
 									<p><span class="hl">Purpose</span></p>
@@ -7267,7 +7267,7 @@ if( GetRaceSkillAdjustment( myRace.id, 32 ) > 10 )
 								</div>
 								<div class="settingsDiv">
 									<p><span class="hl">Example of usage</span></p>
-									<pre><code class="language-javascript">MakeItem( socket, player, 1 );</code></pre>
+									<pre><code class="language-javascript">MakeItem( socket, player, 1, 0x45);</code></pre>
 								</div>
 							</div>
 						</div>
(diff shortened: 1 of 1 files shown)


docs/jsdocs.html
```diff
@@ -7259,15 +7259,15 @@ <h4>January 9th, 2022</h4>
 							<div class="spoiler">
 								<div class="settingsDiv">
 									<p><span class="hl">Prototype</span></p>
-									<p><em>MakeItem( socket, player, itemNum );</em></p>
+									<p><em>MakeItem( socket, player, itemNum, hueNum);</em></p>
```
strattond, line 7262: This isn't actually required (see https://github.com/UOX3DevTeam/UOX3/blob/320bfbe33b2b33b3f7e5621589f735f95dc78ec4/source/SEFunctions.cpp#L474 ) - it's an optional property that defaults to 0.

So perhaps `[, hueNum = 0]` to indicate optionality would be a good option.

docs/jsdocs.html
```diff
@@ -7259,15 +7259,15 @@ <h4>January 9th, 2022</h4>
...
 								</div>
 								<div class="settingsDiv">
 									<p><span class="hl">Purpose</span></p>
 									<p>Player makes the item itemNum. itemNum points to a create.dfn (SECTION ITEM itemNum) entry. Socket can be -1</p>
 								</div>
 								<div class="settingsDiv">
 									<p><span class="hl">Example of usage</span></p>
-									<pre><code class="language-javascript">MakeItem( socket, player, 1 );</code></pre>
+									<pre><code class="language-javascript">MakeItem( socket, player, 1, 0x45);</code></pre>
```
strattond, line 7270: Likewise, whether it's worth adding 2 examples rather than a single one

docs/jsdocs.html
```diff
@@ -7259,15 +7259,15 @@ <h4>January 9th, 2022</h4>
 							<div class="spoiler">
 								<div class="settingsDiv">
 									<p><span class="hl">Prototype</span></p>
-									<p><em>MakeItem( socket, player, itemNum );</em></p>
+									<p><em>MakeItem( socket, player, itemNum, hueNum);</em></p>
 								</div>
 								<div class="settingsDiv">
 									<p><span class="hl">Purpose</span></p>
 									<p>Player makes the item itemNum. itemNum points to a create.dfn (SECTION ITEM itemNum) entry. Socket can be -1</p>
```
strattond, line 7266: What does it mean if socket is -1?  Does that mean it's not a Player but NPC, or just not people being told about it?
Adding some info re hue wouldn't hurt either, particularly if there's an offset from 0 issue

strattond (changes requested):
Just pedantic comments

strattond (approved):
A helpful improvement to the documentation

PR #212, opened 2023-10-15
All UO Halloween

DragonSlayer62:
This All the UO pumpkins that can be double clicked and lit and unlit.


data/js/item/halloweenmasks.js
```diff
@@ -0,0 +1,30 @@
+function onCreateDFN( objMade, objType ) 
+{
+	var randomnamelist2009 = [
```
strattond, line 3: Not a now problem, but ideally there'd be a way to look this up from a list in a DFN somewhere - particularly given namelists.dfn does exist

data/js/item/jackolanterns.js
```diff
@@ -0,0 +1,20 @@
+function onCreateDFN( objMade, objType ) 
+{
+	var randomnamelist2009 = [
```
strattond, line 3: As above

data/js/item/pumpkins.js
```diff
@@ -0,0 +1,324 @@
+// Other settings
+const randomizePumpkinCountdown = false; // If true, add/remove +1/-1 seconds to explosion potion countdowns
+
+function onCreateDFN( objMade, objType ) 
+{
+	var randomnamelist2006 = [
```
strattond, line 6: Now thinking it might be a good time to solve this now ...

strattond (dismissed):
This all looks good to em, but given the hardcoded list of names, I'm wondering if it's worth biting the bullet now and allowing a script to look up a namelist to get a random entry.

Externalises the data into a DFN for easier customisation and reduces risk of fat fingering changes in multiple places?

data/js/item/pumpkins.js
strattond: The other option is to move SetRandomName off the CChar prototype for the JS engine to CBase, and alter the `SetRandomName` function to take a base object rather than char.

It'd be relatively trivial to do

Xoduz (approved):

PR #251, opened 2023-12-18
Bag of Sending

DragonSlayer62:


Xoduz (approved):

data/js/item/bagofsending.js
```diff
@@ -0,0 +1,177 @@
...
+		{
+			socket.SysMessage( GetDictionaryEntry( 19073, socket.language )); // You may only send items from your backpack to your bank box.
+			return false;
+		}
+		else if( bankBox.totalItemCount >= bankBox.maxItems )
+		{
+			socket.SysMessage( GetDictionaryEntry( 19074, socket.language )); // Your bank box is full.
+			return false;
+		}
+		else if( bankBox.weight > bankBox.weightMax )
```
Xoduz, line 103: One thing this does not check, is whether the player's bankbox will be overloaded AFTER the item has been sent to the bank. This should also check if bankBox.weight + ( myTarget.weight * myTarget.amount ) > bankBox.weightMax

data/js/item/bagofsending.js
```diff
@@ -0,0 +1,177 @@
...
+	if( objType == 0 ) 
+	{
+		var Charges = 0;
+		switch( RandomNumber( 0, 2 ))
+		{
+			case 0: Charges = 3; break;
+			case 1: Charges = 6; break;
+			case 2: Charges = 9; break;
+		}
+		objMade.SetTag( "charges", Charges )
```
Xoduz, line 13: Doesn't this mean the bag of sending will always have 3/30 chargers, 6/30 charges or 9/30 charges, since maxCharges is defined (and set) as 30, while the number of charges is always set as 3, 6 or 9?

Also, rename "Charges" variable to "charges"? ;)

Xoduz (changes requested):

Xoduz (approved):

PR #273, opened 2024-04-27
Added Feature Currency Tags

DragonSlayer62:
I think this was always a big miss for emulators is the copper and silver tags for currency just like gold on npcs

Just extra feature for users that like to make rp shards.


Xoduz:
We should not add this, as it's too specific, and too limiting. What if someone wants to add Platinum, or Shillings, or PiecesOfEight? Instead of dedicated DFN tags for each, this is better handled via the existing LOOT tag. Technically this applies to tags like GOLD and BACKPACK too, but we treat them as exceptions for ease of use in the default setup of NPCs.

PR #278, opened 2024-05-01
Lower Requirements Propertie

DragonSlayer62:
Added AOS Lower Requirement propertie.


DragonSlayer62:
After testing this again by itself, seems to be working from the standpoint of AOS. from aos launch

Xoduz (changes requested):
UOGuide mentions that the "Lower Requirement" property also lowers the actual weight of the item:
> It lowers the overall weight of an item by the percentage given.

Might want to work that into CBaseObject::GetWeight(), so the weight returned is reduced by the percentage given in the lower requirement property. That will reduce the effective weight of the weapon, but won't touch the original weight of it.

ALSO, I don't see anywhere reading in the LOWERTATREQ tag from DFNs, only for reading it in from worldfiles.

Lastly, something to consider: It's not just Strength that can be a requirement for equipping items in UOX3. Intelligence and Dexterity properties on items will also determine whether they can be equipped by a player. Should we not support these as well, when we're adding a new property that can lower the requirement of one particular stat (Strength)?

DragonSlayer62:
Per osi it only allows STR i was going off old uo.stratics page from way back machine. for aos

Xoduz (approved):

PR #279, opened 2024-05-05
Spawner Gump

DragonSlayer62:
This gives spawners a gump to edit them, You do not have to use it but very nice if your not wanting to mess with a lot of set commands.


DragonSlayer62:
This has been very helping for the spawners I have placed around my shard without having to remember some commands so this might be ready, but you might want to make some tweaks.

data/js/server/misc/spawnergump.js
```diff
@@ -0,0 +1,248 @@
+function onUseChecked( pUser, iUsed )
+{
+	var socket = pUser.socket;
```
Xoduz, line 3: Add in check here to ensure only GMs can trigger the spawner gump. There might be cases where shards might want a spawner object to be visible to players - example could be a destructible item that will spew out monsters until it is destroyed - and you don't want players to be able to double-click it to edit the spawn properties :)

Xoduz (changes requested):

data/js/server/misc/spawnergump.js
DragonSlayer62: Add isGM check.

data/js/server/misc/spawnergump.js
```diff
@@ -4,7 +4,7 @@ function onUseChecked( pUser, iUsed )
 	socket.tempObj = iUsed;
```
Xoduz, line 4: Shift socket.tempObj = iUsed to inside the if condition, so it only gets set if the conditions are met :)

Xoduz (changes requested):

data/js/server/misc/spawnergump.js
DragonSlayer62: okay

DragonSlayer62: Fixed

Xoduz (approved):

PR #280, opened 2024-05-10
Added ArtifactRarity AOS Property

DragonSlayer62:
ARTIFACTRARITY=#
Artifact Rarity is an item property that is visible on some artifacts and meant to give players an idea of how rare the item is. Items with rarity 1 are supposed to be common while rarity 12 are extremely scarce.


DragonSlayer62:
This works well it's just a new tag to display the artifact status of items it was no longer used after ML. but if your wanting an AOS type shard like most people want this will be something they will want to use. I use it for value on items for players so they can say oh this rare or not rare.

strattond:
The artifact rarity has been added to the *base* object - not to items.

I think it would make sense to attach it to CItem rather than cBaseObject - unless somehow characters can be rare?

DragonSlayer62:
Fixed the artifact tag

source/cItem.cpp
```diff
@@ -542,6 +543,27 @@ auto CItem::SetSpawnerList( bool newValue ) -> void
...
+//o------------------------------------------------------------------------------------------------o
+SI16 CItem::GetArtifactRarity( void ) const
+{
+	return artifactRarity;
+}
+void CItem::SetArtifactRarity( SI16 newValue )
+{
+	artifactRarity = newValue;
+
+	if( CanBeObjType( OT_ITEM ))
```
strattond, line 561: As with the other PR, because we're already in CItem, this should be reducable to just `UpdateRegion();`

We're in Item, so can always be item, and shouldn't need to static cast.

strattond (changes requested):

DragonSlayer62:
fixed

strattond (approved):
Pending testing satisfaction, this looks OK to me

PR #281, opened 2024-05-12
Added Durability Bonus Prop

DragonSlayer62:


DragonSlayer62:
This is ready to go

strattond:
Like the artifact rarity (https://github.com/UOX3DevTeam/UOX3/pull/280), this exposes the stuff to JS via Item, but it is added to the Base Object - this makes more sense to hang off cItem instead, as it doesn't have a conceptual value for a character

source/cItem.cpp
```diff
@@ -156,6 +157,26 @@ auto CItem::GetContSerial( void ) const -> SERIAL
...
+//|	Purpose		-	Gets/Sets the Bonus hp on the object
+//o------------------------------------------------------------------------------------------------o
+SI16 CItem::GetDurabilityHpBonus(void) const
+{
+	return durabilityHpBonus;
+}
+void CItem::SetDurabilityHpBonus(SI16 newValue)
+{
+	durabilityHpBonus = newValue;
+	if (CanBeObjType(OT_ITEM))
```
strattond, line 174: Given this only applies to Items (i.e. CanBeObjType(OT_ITEM) is always true), you can probably remove the IF directly and just call UpdateRegion().

You also shouldn't need to do a static cast, as you're already in CItem at this point.

So this whole if could be reduced to, hopefully, `UpdateRegion();`

strattond (changes requested):

source/cItem.cpp
DragonSlayer62: okay will update

strattond (approved):
Pending testing, this looks OK to me

PR #283, opened 2024-05-14
New Shield Type 107

DragonSlayer62:
Added New Shield Type 107


strattond (commented):
So in theory - looking at the code, I *think* we support the concept of a nested GET.

Would it be worth introducting a `[base_shield]` that inherits from `base_item` and applies type 107, then change the existing shields to get from `base_shield` instead?

It may give you other opportunities for commonising other attributes too (e.g. layer=0x02, decay, etc)

DragonSlayer62:
Okay so I was good at making a type but not sure what you mean with that. sorry still learning the code base

strattond:
> Okay so I was good at making a type but not sure what you mean with that. sorry still learning the code base

So you've updated the dfns to look like this for each shield:

```
[base_bronze_shield]
{
get=base_item
name=bronze shield
id=0x1b72
type=107
decay=1
good=16
layer=0x02
restock=10
str=20
}
```

type=107 and a few other things are all common.  My suggestion would be to change it to something like:

```
[base_shield]
{
get=base_item
type=107
decay=1
good=16
layer=0x02
restock=10
str=20
}

[base_bronze_shield]
{
get=base_shield
name=bronze shield
id=0x1b72
}
```

That way, anything common to a shield is in the `base_shield` section, and each one is just specialised.

I *think* this will work.

DragonSlayer62:
oh, okay I see what you're saying yeah that would work.

DragonSlayer62:
Okay updated

data/dfndata/items/gear/armor/base_armor.dfn
```diff
@@ -1936,218 +1936,156 @@ origin=tol
...
 {
 get=base_item
-name=bronze shield
-id=0x1b72
+type=107
 decay=1
 good=16
 layer=0x02
 restock=10
 str=20
```
strattond, line 1947: This wasn't present on previous definitions - are we sure we want this?

data/dfndata/items/gear/armor/base_armor.dfn
```diff
@@ -1936,218 +1936,156 @@ origin=tol
...
 
 // Metal Kite Shield - Base item with shared DFN tags
 [base_metal_kite_shield]
 {
-get=base_item
+get=base_shield
 name=metal kite shield
 id=0x1b74
-decay=1
-dyeable=1
```
strattond, line 1973: Dyeable was not moved to the base

data/dfndata/items/gear/armor/base_armor.dfn
```diff
@@ -1936,218 +1936,156 @@ origin=tol
...
 [base_wooden_kite_shield]
 {
-get=base_item
+get=base_shield
 name=kite shield
 id=0x1b78
-decay=1
-good=16
-layer=2
-movable=1
```
strattond, line 2012: Movable was not moved to the base

data/dfndata/items/gear/armor/base_armor.dfn
```diff
@@ -1936,218 +1936,156 @@ origin=tol
...
 [base_wooden_shield]
 {
-get=base_item
+get=base_shield
 name=wooden shield
 id=0x1b7a
-decay=1
-good=16
-layer=2
-movable=1
```
strattond, line 2025: Movable was not moved to the base

strattond (changes requested):

data/dfndata/items/gear/armor/base_armor.dfn
DragonSlayer62: you cant dye shields i wander if that was a messup

data/dfndata/items/gear/armor/base_armor.dfn
DragonSlayer62: type 107 is the shield that lets code know what it is.

strattond: No, I mean specifically `str=20`

strattond (approved):

PR #284, opened 2024-05-20
Buff Bar System

DragonSlayer62:
Buffs are positive effects on your character, shown as a green icon.

Debuffs are a negative effect on your character shown as a red icon.

https://uo.com/wiki/ultima-online-wiki/technical/buff-and-debuff-icons/

I only added Healing using buff bar at the moment.

### DIFF
--- a/data/js/server/global.js
+++ b/data/js/server/global.js
@@ -36,6 +36,16 @@ function onLogin( socket, pChar )
     	// Check if "Young" player still meets requirement for being considered young
     	TriggerEvent( 8001, "CheckYoungStatus", socket, pChar, true );
     }
+
+	// Re-adds Buff for disguise kit if player still has time left.
+	var currentTime = GetCurrentClock();
+	var disguiseKitTime = pChar.GetJSTimer(1, 5023);
+	var timeLeft = Math.round((disguiseKitTime - currentTime) / 1000);
+	if (disguiseKitTime != 0)
+	{
+		TriggerEvent(50104, "RemoveBuff", pChar, 1033);
+		TriggerEvent(50104, "AddBuff", pChar, 1033, 1075821, 1075820, timeLeft, "");
+	}
 }
 
 function onLogout( pSock, pChar )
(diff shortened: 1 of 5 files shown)


DragonSlayer62:
This currently works halfway because a lot of things like spells and stuff are in the core so i say we can put this on back burner or it can be added with just the few ones that its working with.

data/js/server/global.js
```diff
@@ -36,6 +36,16 @@ function onLogin( socket, pChar )
     	// Check if "Young" player still meets requirement for being considered young
     	TriggerEvent( 8001, "CheckYoungStatus", socket, pChar, true );
     }
+
+	// Re-adds Buff for disguise kit if player still has time left.
+	var currentTime = GetCurrentClock();
+	var disguiseKitTime = pChar.GetJSTimer( 1, 5023 );
+	var timeLeft = Math.round(( disguiseKitTime - currentTime ) / 1000 );
+	if( disguiseKitTime != 0 )
```
strattond, line 44: I'm concerned with this being `!= 0` - I'd wonder if it should be `> 0` instead, because that implies `disguiseKitTime` is still in the future.

This will also trigger if somehow `disguiseKitTime` is in the past and the timers got backed up.

data/js/skill/healing.js
```diff
@@ -123,17 +123,26 @@ function onCallback1( socket, ourObj )
...
 			var skillNum;
+			var buffIcon;
+			var priCliloc;
+			var scndCliloc;
 			if( IsTargetHealable( ourObj, false ))
 			{
 				// Target can be healed with Healing skill
 				healSkill = mChar.baseskills.healing;
 				skillNum  = 17;
+				buffIcon = 1069;
```
strattond, line 134: I wouldn't be unhappy to know where/how you're finding these magic numbers from.

strattond (changes requested):

data/js/server/global.js
DragonSlayer62: I  see what your talking about

strattond (approved):
These changes look OK to me, from what I can see.  Just need a quick test before merging to make sure it still works after code change @DragonSlayer62

DragonSlayer62:
did one and seems to work way post to be now thats nice.

Xoduz:
[Buff Bar](https://trello.com/c/9On4umAE/257-buff-bar)

PR #285, opened 2024-05-26
Special Moves

DragonSlayer62:
This is the 12 AOS abilites that are in the book on the paperdoll.


DragonSlayer62:
This is complete AOS ready for uox3 It has been on my shard and tested by the players since I built it. This was based on AOS launch.

data/js/combat/special_moves.js
```diff
@@ -0,0 +1,882 @@
...
+		19: { manaAmount: 30 },		// Talon Strike
+		20: { manaAmount: 30 },		// Feint
+		21: { manaAmount: 30 },		// Dual Wield
+		22: { manaAmount: 30 },		// Double shot
+		23: { manaAmount: 30 },		// Armor Peirce // End SE abilities
+		24: { manaAmount: 25 },		// Bladeweave // Start ML abilities
+		25: { manaAmount: 15 },		// Force Arrow
+		26: { manaAmount: 15 },		// Lightning Arrow
+		27: { manaAmount: 25 },		// Psychic Attack
+		28: { manaAmount: 35 },		// Serpent Arrow
```
Xoduz, line 263: This seems to cover more abilities than the original PR description mentioned. Are they all implemented? If not, can comment out the ones that are not covered by this PR?

data/js/combat/special_moves.js
Xoduz, line 1: Please do a pass on this script to convert spaces to tabs and ensure stuff lines up properly :)

data/js/server/global.js
```diff
@@ -37,6 +37,13 @@ function onLogin( socket, pChar )
     	TriggerEvent( 8001, "CheckYoungStatus", socket, pChar, true );
     }
 
+
+	//Attach the special moves Book
+	if( !pChar.npc && !pChar.HasScriptTrigger( 7001 ))
+	{
+		pChar.AddScriptTrigger( 7001 );
+  }
+
```
Xoduz, line 46: Add a new setting in UOX.INI: ENABLESPECIALMOVES=#
Make it a bitset style setting, where 0x1 enables LBR special moves, 0x2 enables AOS ones, 0x4 SE ones, 0x8 ML ones and 0x10 SA ones.

Then add a check here to see if the setting is higher than 0 before attaching the script.

data/js/server/global.js
```diff
@@ -81,6 +88,12 @@ function onCreatePlayer( pChar )
 
 		TriggerEvent( 8001, "GiveYoungPlayerItems", pChar.socket, pChar );
 	}
+
+	//Attach the special moves Book
+	if( !pChar.npc && !pChar.HasScriptTrigger( 7001 ))
+	{
+		pChar.AddScriptTrigger( 7001 );
+	}
```
Xoduz, line 96: See above comment about adding uox.ini setting and checking for it here before attaching script to player.

data/js/combat/special_moves.js
```diff
@@ -0,0 +1,882 @@
+function onSpecialMove( pUser, abilityID )
```
Xoduz, line 1: I don't see anywhere in the PR that references or calls upon onSpecialMove. What am I missing? How does it get called?

Xoduz (changes requested):

data/js/server/global.js
DragonSlayer62: There was no special moves book in LBR just AOS and up.

data/js/combat/special_moves.js
DragonSlayer62: I added all thge abilties for future expansions reason for the restrictedAbilities

data/js/combat/special_moves.js
DragonSlayer62: that is the script being attached to the char the packet is calling this when you click the book something like quest button.

Xoduz: Oops, that's true :) Thanks

data/js/server/global.js
Xoduz: You're right, I was thinking of the original crushing blow/paralyzing blow/concussion blow from UOR, but we already added support for those previously. We should still do the setting for AOS/SE/ML/SA special moves though :)

Xoduz (approved):

PR #290, opened 2024-06-14
Hit/Defense Chance Properties

DragonSlayer62:
Hit Chance Increase - Increases the player's chance to hit a target with wrestling, melee and ranged weapons. (The maximum effect you can get from this property is capped at 45%)

Defense Chance Increase - Increases the wearer's chance that his opponents' swings (or arrows/bolts) will miss. (The maximum effect you can get from this property is capped at 45%)


DragonSlayer62:
Most the code was already in place for this just needed some extra to get it to work. This is ready to go.

PR #291, opened 2024-06-16
Swing Speed Increase

DragonSlayer62:


source/cServerData.cpp
Xoduz, line 1: Make sure to set a default value for SwingSpeedIncreaseCap() setting via CServerData::ResetDefaults(). Can also mention the default value in changelog entry.

source/combat.cpp
Xoduz, line 1: The changes in CHandleCombat::GetCombatTimeout() to calculate swing delay based on era won't work as intended, because the weapons in the different era-specific DFNs has the swingspeed formulas of each specific era baked into the SPD values (determined via a spreadsheet!). This allows applying era-appropriate swing speeds to weapons from _future_ eras in _older_ eras while keeping the balance between weapons relatively intact, so someone can do an LBR-focused shard but allow weapons from future eras without upsetting the balance.

Please change this to just insert the swingspeed bonus increase into the default formula, i.e.
`getDelay = baseValue / ( getDelay * getOffset * speedFactor ) / globalAttackSpeed;`

source/Changelog.txt
```diff
@@ -1,3 +1,13 @@
+16/06/2024 - Dragon Slayer/Xuri
+	Added new DFN tags for Items:
```
Xoduz, line 2: This seems to be supported for both Items _and_ Characters?

Xoduz (changes requested):

source/cServerData.cpp
DragonSlayer62: it is already set at 60 for when ini loads.

source/combat.cpp
DragonSlayer62: like this

getDelay = baseValue / ( getDelay * getOffset * speedBonus ) / globalAttackSpeed;

source/Changelog.txt
DragonSlayer62: yes ill update that

DragonSlayer62:
fixed

source/combat.cpp
Xoduz: Keep this:
`SI32 baseValue = 15000;`
and this one for fleeing targets:
`baseValue = 10000; `

Keep the speedfactor stuff, but replace the if/else stuff at the bottom with just that one line:
`getDelay = baseValue / ( getDelay * getOffset * speedFactor ) / globalAttackSpeed;`

source/cServerData.cpp
Xoduz: I don't see that in this PR? There's usually a line per setting in cServerData that sets the default, but that doesn't seem to be present for the swingspeedincreasecap.

source/combat.cpp
Xoduz: I have mashed potato for brains. Ignore this.

Xoduz (approved):

PR #294, opened 2024-07-07
Requried Free Hand

DragonSlayer62:
Added Requirement to drink potion hand has to be free, This can be disabled within the potion.js file by switching the reqfreehand to false.


source/Changelog.txt
```diff
@@ -1,3 +1,7 @@
+
+7/07/2024 - Dragon Slayer
+	Added Required Free Hand (js potion).
```
Xoduz, line 3: Elaborate. You wrote it so nicely in the PR description, do the same for changelog entries! :) These are the entries that are used to list out changes with releases, so people can scroll through and see what fixed/additions/changes were made. "Added Required Free Hand" doesn't say a lot, but your PR description does (just update "reqfreehand" to match the variable name in the script)

> Added Requirement to drink potion hand has to be free, This can be disabled within the potion.js file by switching the reqfreehand to false.

Xoduz (changes requested):

source/Changelog.txt
DragonSlayer62: okay

Xoduz (approved):

PR #301, opened 2024-08-14
Added ML Quivers (dfn clothing)

uogonemetal:


data/js/item/equip_effects/updatequivercount.js
```diff
@@ -0,0 +1,52 @@
+function onSwing(iSwung, pAttacker, pSwingAt)
+{
+	UpdateQuiverCount(pAttacker);
+	return true;
+}
+
+function onAttack( pAttacker, pDefender, hitStatus, hitLoc, damageDealt )
```
Xoduz, line 7: This has a dependency on https://github.com/UOX3DevTeam/UOX3/pull/300, which should be merged first.

data/js/item/archerybutte.js
```diff
@@ -140,7 +143,19 @@ function onUseUnChecked( pUser, iUsed )
...
+					var totalAmmoCount = 0;
+					for( var mItem = cloakItem.FirstItem(); !cloakItem.FinishedItems(); mItem = cloakItem.NextItem() )
+					{
+						if( ammoType == "Arrow" && mItem.id == 0x0f3f )
+						{
+							totalAmmoCount += mItem.amount;
+							if( mItem.amount > 0 )
+								mItem.amount -= 1;
+							else
+								mItem.Delete();
```
Xoduz, line 155: It seems like this part might be double-dipping on arrow removal. It both iterates through the quiver and reduces arrow count by 1/deletes item if only 1 left, AND it continues on to the part below which consumes an arrow via pUser.UseResource(). Since UseResource() already iterates through all containers in player's pack until it finds what it needs, remove the manual .amount change and .Delete().

data/js/item/archerybutte.js
```diff
@@ -156,7 +171,19 @@ function onUseUnChecked( pUser, iUsed )
...
+					var totalAmmoCount = 0;
+					for( var mItem = cloakItem.FirstItem(); !cloakItem.FinishedItems(); mItem = cloakItem.NextItem() )
+					{
+						if( ammoType == "Bolt" && mItem.id == 0x1bfb )
+						{
+							totalAmmoCount += mItem.amount;
+							if( mItem.amount > 0 )
+								mItem.amount -= 1;
+							else
+								mItem.Delete();
```
Xoduz, line 183: See previous comment about double-dipping on arrow removal. Same applies here.

data/dfndata/items/gear/clothing/clothing.dfn
```diff
@@ -489,3 +489,31 @@ id=0x1ffe
...
+type=1
+weight=200
+weightmax=50000
+maxitems=500
+value=400 200
+value=2 1
+layer=0x14
+script=5061
+custominttag=WeightReduction 50
+}
```
Xoduz, line 507: Add origin=ml tag please :)

Xoduz (changes requested):

data/js/item/archerybutte.js
DragonSlayer62: Reason i doing it that way because it wasnt looking at the quiver and taking the arrow out it was deleting bolts and arrows and this was causing the quiver to break if i remove that it breaks the quiver with archery buttes I know there is a method to my madness with this i promise.

data/js/item/equip_effects/updatequivercount.js
DragonSlayer62: done

data/js/item/archerybutte.js
Xoduz, line 1: Yeah, this is a no-go. Too many unnecessary changes, and entire portions of the script have been removed with this rework (splitting arrows in bullseye, turning towards butte, etc). I left comments in discord PM, please continue discussion there to address the issues I raised, thanks.

Xoduz (changes requested):

PR #304, opened 2024-08-31
Fixed Height check

DragonSlayer62:
Fixed Height Check for clicking on dynamic items


Xoduz (changes requested):
All this needs is a changelog entry that spells out the gist of the changes :)

Xoduz (approved):

PR #306, opened 2024-09-28
Function Fixes

DragonSlayer62:
functions for checking boats.
Also magic.js had IsBoat spelling wrong been corrected


Xoduz (changes requested):
What are these changes for? What can people do in scripts after these changes that they couldn't before? Changelog plz :)

Xoduz (approved):

PR #307, opened 2024-10-10
puzzle chest

DragonSlayer62:
This was osi Puzzle chest for kahldun.


Xoduz (approved):

Xoduz (changes requested):
I previously approved this, but after looking at it again, I think it needs a changelog update that explains what was added, and how it can be potentially used. However, if there's no way to actually make use of it right now, we may want to delay merging this until it has some more context.

DragonSlayer62:
Added changelog with description what it does

Xoduz (changes requested):
As part of the onCreateDFN event, these puzzle chests should be hidden as soon as they spawn. Then the detect hidden skill should be updated so it can be used to uncover these. Maybe detecting one should trigger a timer on the chest that will re-hide it if it hasn't been opened in X minutes, to reset the experience for the next player? :)

Also, since we don't have a setup for Khaldun dungeon spawns etc yet, the very least we should do before adding these to UOX3 is include them in the 'add menu as something admins/GMs can find/add and consider spawning elsewhere themselves, so it doesn't just become a "lost" feature that no one knows exists.

DragonSlayer62:
I added the chest to the addmenu.

DragonSlayer62:
After looking on way back machine this chest was never hidden that had to be a update they added since then because only current uoguide says there hidden.

Xoduz (approved):

PR #309, opened 2024-10-12
Mini Houses Added in AOS

DragonSlayer62:
All the Mini House Add-Ons

### DIFF
--- a/data/dfndata/house/house_aos.dfn
+++ b/data/dfndata/house/house_aos.dfn
@@ -1,0 +1,290 @@
+// House Addon - Mini House: Stone and Plaster
+[HOUSE 500]
+{
+ID=0x14F0
+SPACEX=1
+SPACEY=1
+CHARX=0
+CHARY=0
+CHARZ=0
+HOUSE_DEED=stoneandplastermini
+HOUSE_ITEM=1500
+}
+
+// House Addon - Mini House: Field Stone
+[HOUSE 501]
+{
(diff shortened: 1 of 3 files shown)


data/dfndata/house/house_aos.dfn
strattond, line 1: Given a big chunk of these have a common ID, SPACEX, SPACEY and CHARXYZ, is it worth adding a `[baseminihouse]` and using GET to inherit the common stuff?

strattond (changes requested):
Minor suggestion, but looks reasonable to me.  Haven't done a full box and dice cross-validation of content.

data/dfndata/house/house_aos.dfn
Xoduz: Would be useful, but don't think that is supported atm, because the loading of the house.dfn data is handled directly in the "BuildHouse" function in house.cpp, and doesn't currently support a GET tag. If we want to support that, need to separate out the dfn processing from that function, I think.

source/Changelog.txt
```diff
@@ -1,3 +1,6 @@
+12/10/2024 - Dragon Slayer
+	Added Mini-House Addons from ML Expansion.
```
Xoduz, line 2: From AoS Expansion, not ML :)

Xoduz (changes requested):

Xoduz (approved):

PR #310, opened 2024-10-15
CheckInstaLog  Added

DragonSlayer62:
Added the CheckInstaLog function to check if a specific location is within an instant logout zone (SEFunctions.cpp).


docs/jsdocs.html
Xoduz, line 1: Same comment as in another PR - please make sure only the relevant changes to the document is submitted (with tabs, not spaces ;), and not the entire file full of changes (tabs to spaces, in this case).

Xoduz (changes requested):

DragonSlayer62:
Fixed Doc

Xoduz (approved):

PR #312, opened 2024-10-20
Added SetRandomColor

DragonSlayer62:
Applies a random color from specified colorlist to character or item


docs/jsdocs.html
Xoduz, line 1: This seems like a pretty drastic change just to add something new to the file. Looks like all tabs have been converted to spaces in the file, which is understood as an update of every single line in the entire file. Please address :)

Xoduz (changes requested):

docs/jsdocs.html
Xoduz: 👆

DragonSlayer62:
Fixed Doc

Xoduz (approved):

PR #316, opened 2024-12-15
Adding Ressurrect Penalty

uogonemetal:
Script that controls any type penalty you want on the char after resurrecting.


Xoduz (changes requested):
Feels like something is missing here. At the moment it's a standalone script that's not attached to anything, does not trigger from anything, and with no way to enable/disable.

I suggest adding this as a setting to UOX.INI - RESURRECTIONPENALTIES=0/1 (or some such), and then use global.js to attach the script to players who die, via the existing onDeath event there, if the setting is enabled in INI.

Then add onResurrect( pAlive ) to your script and trigger the resurrect penalties from there (again, if INI setting is enabled).

DragonSlayer62:
This uses the karmalock pr that has the shrine its attached to it.  you do not want to attach this to players at all

Xoduz (approved):

PR #320, opened 2024-12-29
Remove superfluous parameter from SE_FindRootContainer

DragonSlayer62:
fixed the check for 1 argument, not 2.

### DIFF
--- a/source/SEFunctions.cpp
+++ b/source/SEFunctions.cpp
@@ -2028,9 +2028,9 @@ JSBool SE_GetPackOwner( JSContext *cx, [[maybe_unused]] JSObject *obj, uintN arg
 //o------------------------------------------------------------------------------------------------o
 JSBool SE_FindRootContainer( JSContext *cx, [[maybe_unused]] JSObject *obj, uintN argc, jsval *argv, jsval *rval )
 {
-	if( argc != 2 )
+	if( argc != 1 )
 	{
-		ScriptError( cx, "FindRootContainer: Invalid number of arguments (takes 2)" );
+		ScriptError( cx, "FindRootContainer: Invalid number of arguments (takes 1 - item or serial)" );
 		return JS_FALSE;
 	}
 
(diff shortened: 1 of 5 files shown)


source/SEFunctions.cpp
```diff
@@ -2028,9 +2028,9 @@ JSBool SE_GetPackOwner( JSContext *cx, [[maybe_unused]] JSObject *obj, uintN arg
 //o------------------------------------------------------------------------------------------------o
 JSBool SE_FindRootContainer( JSContext *cx, [[maybe_unused]] JSObject *obj, uintN argc, jsval *argv, jsval *rval )
 {
-	if( argc != 2 )
+	if( argc != 1 )
 	{
-		ScriptError( cx, "FindRootContainer: Invalid number of arguments (takes 2)" );
+		ScriptError( cx, "FindRootContainer: Invalid number of arguments (takes 1 - item or serial)" );
 		return JS_FALSE;
 	}
```
strattond, line 2036: The mType following still refers to the old, second parameter - I don't know what behaviour is expected here.

If it's going to be an item or serial, do we need to discriminate on if JSVAL_IS_OBJECT instead?

i.e. something like

```c++
if( JSVAL_IS_OBJECT( argv[0] ) ) // item
{
...
}
else
{
...
}

strattond (changes requested):
Changelog please.

Also, you've removed a variable but still reference it in the code.  This can lead to unexpected behaviour.

source/SEFunctions.cpp
DragonSlayer62: I have to ask xuri he was the one that helped me with this one

Xoduz: I must not have been conscious when I did that. It clearly needs the second parameter to determine whether it's an item (second param is 0) or a serial (second param is 1, or... anything other than 0 heh). Think can scrap this PR.

PR #321, opened 2025-01-01
New Haven Integration and Music Updates for region scp

DragonSlayer62:
Added New Haven for trammel
Added New Haven Music
Added new player quest training areas
Fixed Britains music
marked old regions file as old because contains old haven


data/dfndata/regions/regions.dfn_old
strattond, line 1: Unless this is going to be used, I would remove this from the PR - it's not active by this extension

strattond (changes requested):
Changelog please.

Also, a more meaningful title would be good

data/dfndata/regions/regions.dfn_old
DragonSlayer62: i left it in like we did the maps file just incase some one wanted to use it, We could just post the information on the boards and remove it. if that be better.

source/Changelog.txt
```diff
@@ -1,3 +1,10 @@
+22/01/2025 - Dragon Slayer
+	Added New Haven for trammel.
```
Xoduz, line 2: Added New Haven / New Haven music where? :) if in regions.dfn, please specify 🙏

DragonSlayer62:
udpated it now says (regions.dfn)

data/dfndata/regions/regions.dfn_old
Xoduz: I'll reiterate what I wrote in a discord PM:

- Rename the file to "regions_old_haven_only.bak"
- Yoink out anything that's not old haven from the file
- Leave a comment at the top of the file stating that this file defines the town region for old haven, which only exists on Trammel facet in client versions 5.0.0b to 7.0.8.2 with mapdiffs enabled, and if anyone wants to use it, copy it to the main regions.dfn file and overwrite the New Haven region (57), while also removing the New Haven specific shop-regions (178 to 200)

Xoduz (approved):

PR #322, opened 2025-01-04
Script Trigger Assignment and Memory Management.

DragonSlayer62:
Unequip fix
fixed spelling mistake in script


strattond:
@DragonSlayer62 changelog + more meaningful title please

source/cChar.cpp
```diff
@@ -2991,6 +2991,9 @@ bool CChar::TakeOffItem( ItemLayers Layer )
 			}
 		}
 
+		scriptTriggers.clear();
```
strattond, line 2994: I don't understand what this is trying to achieve.  I get what it does, but not the goal.

`clear()` and `shrink_to_fit()` are going to be superfluous because we're just replacing the content by calling GetScriptTriggers again.

But ... why are we calling GetScriptTriggers again?  Is it possible that a previous script trigger has altered all the script triggers for the item of interest?

Need a better understanding of what this one is for.

Xoduz: Hm I wonder if there's something missing here, or intended for a different PR :) I remember discussing this with @DragonSlayer62 and concluding that an onUnEquip event should be added for the character as well, right after the item's one - which would also explain the clearing/shrinking/getting of scripttriggers again from item for onUnEquip just below. Though in that case it might be just as well to create two variables - one for item's scripts, and one for character's? The same would also be the case for the WearItem function just above.

The general idea of clearing & shrinking of the vector before reassigning is possibly something that stems from a potentially misguided attempt on my part to follow best practices randomly uncovered on stack overflow many years ago, and can also be seen in other places in the codebase 😅

source/cChar.cpp
Xoduz, line 1: Ensure usage of tabs instead of spaces and correct indentation in both WearItem and TakeOffitem functions
image
image

Xoduz (changes requested):

DragonSlayer62:
Fixed the Tabs in the file you IDed

Xoduz (approved):

PR #326, opened 2025-01-17
Durability HP Bonus support

DragonSlayer62:
fixed pr


source/items.cpp
```diff
@@ -1194,11 +1195,22 @@ CItem * cItem::CreateBaseScriptItem( CItem *mCont, std::string ourItem, const UI
...
+		{
+			iCreated->SetHP( iCreated->GetHP() + durabilityHpBonus );
+		}
+
 		// If maxHP has not been defined for a new item, set it to the same value as HP
 		if( !iCreated->GetMaxHP() && iCreated->GetHP() )
 		{
 			iCreated->SetMaxHP( iCreated->GetHP() );
 		}
+		else
```
strattond, line 1210: I'm not sure this does what you want...
The first if will set MaxHP == HP if it hasn't been set.  Cool.
But the second increases it based on durability HP bonus, but only if it has both a MaxHP and HP already set.

I suspect this `else` should become `if( durabilityHpBonus > 0 )`

DragonSlayer62: I fixed it also I changed the percent.

strattond (approved):
Seems reasonable to me, no obvious problems.

source/Changelog.txt
```diff
@@ -1,6 +1,10 @@
 13/05/2024 - Dragon Slayer
 	Added New Shield Type 107 so shield ID's no longer have to be in hard code for shields to work properly
 
+11/05/2024 - Dragon Slayer/Xuri/Maarc
+	Added Support for new AOS property tag.
+	-DURABILITYHPBONUS - Add durability (aka Health) to the hp and max hp as a bonus
```
Xoduz, line 6: Just DURABILITYHPBONUS as a standalone tag, or with a value? And if a value, does it represent additional hitpoints or a percentage multiplier?

PR #327, opened 2025-01-21
Damage Adjustment should not use any skill

DragonSlayer62:
Neg Damage Does not take into account the attacker skill no longer. Only the Armor perc


strattond:
@DragonSlayer62 changelog please

strattond (approved):
Pending @Xoduz approval, this works for me

source/Changelog.txt
```diff
@@ -1,3 +1,6 @@
+21/01/2025 - Dragon Slayer
+	Removed the skill check for damage being adjusted at the end. (combat.cpp)
```
Xoduz, line 2: Even though skills are involved, this is not a skill check :) Just say you fixed an issue where damage dealt in combat was incorrectly modified while applying defense modifiers :P

DragonSlayer62: Fixed

PR #343, opened 2025-02-23
Damage Increase

DragonSlayer62:
Damage Increase Property increase damage by a percentage


source/UOXJSPropertyFuncs.cpp
```diff
@@ -2027,6 +2028,7 @@ JSBool CCharacterProps_getProperty( JSContext *cx, JSObject *obj, jsval id, jsva
 			case CCP_SPATTACK:		*vp = INT_TO_JSVAL( gPriv->GetSpAttack() );					break;
 			case CCP_SPDELAY:		*vp = INT_TO_JSVAL( gPriv->GetSpDelay() );					break;
 			case CCP_SWINGSPEEDINCREASE:	*vp = INT_TO_JSVAL( gPriv->GetSwingSpeedIncrease() );		break;
+			case CCP_DAMAGEINCREASE:*vp = BOOLEAN_TO_JSVAL( gPriv->GetDamageIncrease() );		break;
```
Xoduz, line 2031: This appears to be incorrect. Should be INT_TO_JSVAL, no?

source/cBaseObject.cpp
```diff
@@ -2091,6 +2124,7 @@ bool CBaseObject::HandleLine( std::string &UTag, std::string &data )
 					SetHitChance( static_cast<SI16>( std::stoul( oldstrutil::trim( oldstrutil::removeTrailing( csecs[0], "//" )), nullptr, 0 )));
 					SetDefenseChance( static_cast<SI16>( std::stoul( oldstrutil::trim( oldstrutil::removeTrailing( csecs[1], "//" )), nullptr, 0 )));
 					SetSwingSpeedIncrease( static_cast<SI16>( std::stoul( oldstrutil::trim( oldstrutil::removeTrailing( csecs[2], "//" )), nullptr, 0 )));
+					SetDamageIncrease( static_cast<SI16>( std::stoul( oldstrutil::trim( oldstrutil::removeTrailing( csecs[3], "//" )), nullptr, 0 )));
```
Xoduz, line 2127: This might trigger a crash if trying to load a set of worldfiles that does not have the damageIncrease data in the EXTPROPCOMMON worldfile tag already. Enclose in if( csecs.size() >= 4 ):
```suggestion
					if( csecs.size() >= 4 )
					{
						SetDamageIncrease( static_cast<SI16>( std::stoul( oldstrutil::trim( oldstrutil::removeTrailing( csecs[3], "//" )), nullptr, 0 )));
					}
```

source/combat.cpp
```diff
@@ -2117,6 +2117,8 @@ SI16 CHandleCombat::ApplyDamageBonuses( WeatherType damageType, CChar *mChar, CC
 	R32 multiplier = 1;
 	R32 damage = 0;
 	SI32 RaceDamage = 0;
+	// Add Damage Increase (DI) from items, buffs, etc.
+	SI16 damageIncreasePercent = mChar->GetDamageIncrease(); // Total DI from items, buffs, etc.
```
Xoduz, line 2121: This only applies to PHYSICAL damage currently, so move it into the switch section, under case PHYSICAL: (just before the other changes there)

Xoduz (changes requested):

Xoduz (approved):

PR #363, opened 2025-04-22
Tooltip sorting

Xoduz:
Tooltip entries for Characters/Items now have a sorting order associated with them, defined in code via a "sortOrder" parameter in the ToolTipEntry For custom tooltips in scripts, sort order can be set via the temporary tag. Example:
	iUsed.SetTempTag( "tooltipSortOrder", 1 ); // sort tooltip after name (sortOrder 0), but before maker's mark (sortOrder 5).
This information and a list of the default hard-coded sort order can also be found in the updated UOX3 Documentation, under Features > Object Tooltips


DragonSlayer62 (commented):
Looks good after testing

DragonSlayer62 (approved):

PR #374, opened 2025-05-10
Healing Overhaul, fixes for mana regen, skill caps

Xoduz:
- Updated global script (js/server/global.js) to make use of onUseBandageMacro() to enable usage of bandage macros in client
- Overhaul of Healing skill (js/skill/healing.js and js/skill/healing_slip.js) with fixes and improved era support. Script now includes some options that can be modified at top of script to override era-specific behavior. Summary of changes:
  - Healing actions are now difficulty based:
    - Resurrect chance scales from 25% (min 80.0 primary/secondary skill) to 65% at avg 100.0 skill to 100% at avg 120.0 skill
    - Cure chance scales from 80% (min 60.0 primary/secondary skill) to 92% at avg 100.0 skill and 100% at avg 120.0 skill, and is reduced by 10% per level of poison on target
    - Upper difficulty for Healing skill checks now changes based on % loss of health on target and based on its maxhp, so to continue gaining skills from healing, one needs to heal targets with higher and higher % of their health lost
    - For T2A core shard era, healing can only bring player to ~65.0 skill, curing can bring player from 60.0 to ~85.0 skill, and resurrection can bring player from 80.0 to 100+ skill.
    - For UOR and beyond, healing damage can bring player all the way to GM skill and beyond, without need to cure/resurrect
  - Base amount of damage healed with bandages in SE era and beyond adjusted to 36-60, compared to 43-80 for AoS and below
  - In HS era and beyond, min 80.0 primary and 80.0 secondary healing skill will now trigger attempt at curing poison/bleed at half normal heal duration (but with lower chance of success). If it succeeds, amount of health healed is reduced proportional to strength of poison cured. If it fails, another cure attempt is made after the heal.
  - Mortal Strike now blocks heal attempts on a target, but not attempts to cure poison/bleed
  - Fingers slipping due to being attacked while bandaging now has a higher threshold to take place for SE era and beyond, and scales further based on dexterity in HS era and beyond - while AOS era and below has a very low threshold for this
- Updated .CheckSkill() Character JS Method to support another parameter which can force the result of a skillcheck (-1 = failure, 0 = use skillcheck result, 1 = success):
  - .CheckSkill( skillNum, minSkill, maxSkill[, isCraftSkill, forceResult ] )
- Updated onSkillCheck() JS Event with another parameter, with the value of the optional forceResult parameter in CheckSkill()
  - onSkillCheck( pUser, skillID, lowSkill, highSkill, forceResult )
- Fixed an issue where skill caps for player characters were not correctly initialized, resulting in unpredictable behavior
- Fixed an issue with mana regen timer for MANAREGENMODE=1 (LBR and below) where timer was never adjusted
- Fix for pets getting stuck staring at potential targets in combat while simultaneously ignoring them
- The damage of a weapon can now also be modified via the 'set damage # command. Supported syntax:
  - 'set damage # // sets both lodamage and hidamage to value of #
  - 'set damage # # // sets lowdamage to first value, hidamage to second value
- Added two new Admin commands to help address issues with player characters. Use sparingly!
  - 'resetskillcaps // Resets ALL individual skillcaps for ALL player characters back to default
  - 'resetskillusage // Resets skill usage status of all skills for ALL player characters to false, to unblock their use

### DIFF
--- a/data/dictionaries/dictionary.CSY
+++ b/data/dictionaries/dictionary.CSY
@@ -3739,6 +3739,15 @@
 8213=ID sekce z DFNs objekt pochází z
 8214=Dá se předmět ukrást? 0 = nelze ukrást, 1 = lze ukrást, 2 = lze ukrást vzácně
 8215=Neplatný cíl. Vyberte platnou položku nebo NPC
+8251=Vymazání NPC...
+8252=Vymazáno %i NPC
+8253=Vyberte NPC, které chcete přesunout na dálku pomocí xrun:
+8254=Vyberte NPC, které chcete přesunout na dálku pomocí xwalk:
+8255=Vyberte NPC, které chcete otočit na dálku pomocí xturn:
+8256=Vyberte cíl pohybu (%s):
+8257=Musíte vybrat NPC!
+8258=Ošetřil jsi ránu a zastavil krvácení
+8259=Krvácející rány se zahojily, už nekrvácíš!
 // [8300-8499] Tweak menu - Character property descriptions
 8300=Číslo účtu přidružené k hráči. Řízeno serverem (pouze pro čtení)
 8301=Typ umělé inteligence NPC
--- a/data/dictionaries/dictionary.ENG
+++ b/data/dictionaries/dictionary.ENG
@@ -3741,6 +3741,15 @@
 8213=Section ID from DFNs object originated from
 8214=Is item stealable? 0 = not stealable, 1 = stealable, 2 = stealable rare
 8215=Invalid target. Please select a valid item or NPC
+8251=Wiping NPCs...
+8252=Wiped %i NPCs
+8253=Select NPC to move remotely using xrun:
+8254=Select NPC to move remotely using xwalk:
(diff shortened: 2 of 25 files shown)


DragonSlayer62 (approved):
everything seems good from a standpoint healing seems more on with current osi after I did testing with new and old era

DragonSlayer62 (approved):
after testing and adjusting the skill in the dfn and changing it i was able to say gains and stuff work alot better.

[SKILL 17]
{
NAME=HEALING
STR=60
DEX=60
INT=80
SKILLPOINT=0,100,100
SKILLPOINT=300,75,50
SKILLPOINT=500,55,30
SKILLPOINT=700,35,15
SKILLPOINT=800,25,10
SKILLPOINT=900,15,5
SKILLPOINT=950,8,2
SKILLPOINT=980,5,1
SKILLPOINT=990,2,0
SKILLPOINT=1000,0,0
}

update i did to my skills.dfn

this also made it easier for me to test the entire healing script too.

PR #376, opened 2025-05-14
Fix for skill calculation and int contribution to skill

Xoduz:
- Fixed an issue where integer division and rounding of values could lead to skill gains not being visible to player
- Fixed an issue where intelligence stat would not contribute to skill points based on skills.dfn setup the same way as dex/str


DragonSlayer62 (approved):
Okay I'm not happy I enjoyed my 0.2 points of skill, just kidding I would have never noticed until you said something, But I guess this works I won't drag your fix down I promise.

PR #399, opened 2025-07-20
Updates to Visual Studio build configuration

Xoduz:
- [FIX] Fixed an issue where animal trainer shoplist was referring to an incorrect sectionID for desert ostards
- [FIX] Restored support for VS2017
- [UPD] Updated VS solutions/projects to once again build and link Spidermonkey using a static library instead of producing a DLL
- [UPD] Turned console warning about string properties into a log-action that writes directly to warning.log instead


DragonSlayer62 (approved):
I Approve this changes. Make UOX3 great

PR #402, opened 2025-08-04
Added way to define creatures as pack animals

Xoduz:
- [ADD} Added new Creatures DFN tag that can mark a creature as a pack animal: PACKANIMAL
- [ADD] New Character JS Property to check if NPC is a pack animal: .isPackAnimal


DragonSlayer62 (approved):
Tested and Working.

PR #404, opened 2025-08-04
House decay

DragonSlayer62:
- [ADD] Added House Decay System, which automatically marks and collapses unattended houses over time. Decay stages and timers are configurable via the uox.ini file.


data/js/item/doors.js
```diff
@@ -193,6 +193,16 @@ function onUseChecked( pUser, iUsed )
...
 			}
 		}
+
+		// Refresh house decay timer if the door is opened
+		if( iUsed.multi && ( !iUsed.multi.GetTag( "InDanger" )))
+		{
+			const decayStageLikeNewMins = GetServerSetting( "DecayStageLikeNewMins" );
+			const decayLikeNewMS = decayStageLikeNewMins * 60 * 1000;
+
+			iUsed.multi.KillTimers();
```
Xoduz: This is dangerous, as it will kill _ALL_ timers associated with the Multi - even ones not covered by these scripts. Safer to use .KillJSTimer() Method to target specific timers. Example:

``iUsed.multi.KillJSTimer( 0, 4500 ); // timerID, scriptID``

data/js/server/house/houseSign.js
```diff
@@ -471,6 +471,24 @@ function onGumpPress( pSocket, pButton, gumpData )
 				TriggerEvent( 15002, "DeclareHousePrivate", pSocket, iMulti );
 			}
 			break;
+		case 60: // Turn off Grandfathered status
+			if( pUser.isGM || iMulti.IsOwner( pUser ))
```
Xoduz: Should owner of house really be able to turn grandfather status of the house on/off at will? Shouldn't this be a GM only thing?

Xoduz (changes requested):

data/js/server/house/houseSign.js
DragonSlayer62: you right should only be a gm thing

data/js/item/doors.js
DragonSlayer62: okay i will change this

DragonSlayer62: I will have to figure a way to do this because the timer isnt just 1 ID i will have to know what timer ID its on maybe have to save the timer ID in a tag and call it back so this knows what one to kill.

data/js/server/house/house.js
```diff
@@ -74,6 +88,14 @@ function onHouseCommand( pSocket, iMulti, cmdID )
 /** @type { ( left: Multi, leaving: BaseObject ) => boolean } */
 function onEntrance( iMulti, charEntering, objType )
 {
+	//Start Decay timer if decay is enabled or init is not true.
+	if( !iMulti.GetTag( "init" ))
```
Xoduz: "init" is too generic name for a tag. Other scripts could also be on houses that require initialization, and would need to do their own checking vs such a tag. Make this more unique to the feature, like... "houseDecayInit" for example.

data/js/server/house/house.js
```diff
@@ -202,6 +224,177 @@ function onMultiLogout( iMulti, cPlayer )
...
+	//Skip decay if this house is marked Grandfathered and its turned on
+	if( houseGrandFatheredSystem == 1 && iMulti.GetTag( "Grandfathered" ))
+	{
+		var GetStage = parseInt(iMulti.GetTag("decayStage"));
+		iMulti.KillJSTimer( GetStage, 15000 ); // timerID, scriptID
+		//iMulti.KillTimers();
+		return;
+	}
+
+	var choseDays = Math.random() < 0.5 ? decayLowMS : decayHiMS;// Random 2 to 3 days timer.
```
Xoduz: What does "choseDays" mean? Can you give this a more relevant name, like.. if it's just a variable for randomizing the amount of days until the next decay stage, rename it to "rndDecayTime" or "rndDecayStageTime" or something.

source/cServerData.cpp
```diff
@@ -986,6 +996,13 @@ auto CServerData::ResetDefaults() -> void
 	KeylessCoOwnerAccess( true );
 	KeylessFriendAccess( true );
 	KeylessGuestAccess( false );
+	HouseDecay( false );
```
Xoduz: These define the default values for house decay. Why is the system disabled by default? Was there no house decay in LBR era?

Also, please include an updated uox.ini file that include the new house-decay related settings. Start up a fresh server, do a worldsave, submit updated uox.ini file minus the changes done to get server to start (path to UO client).

Xoduz (changes requested):

Xoduz (approved):

PR #414, opened 2025-08-13
Initial landing of TypeScript bindings of exposed APIs and classes

strattond:
- [ADD] Typescript definitions for events, methods and exposed classes from the core
  - NOTE: This is *incomplete* and further revisions may be required.  Additionally, it does NOT force you to use it - if you code the existing way, it'll still work.  This just gives you sanity on scripting.
- [ADD] Add PowerShell script to create new TS file from core JSON updates
- [ADD] Add PowerShell script to update or add JSDoc for appropriate functions in existing scripts
- [UPD] Existing scripts have JSDoc annotations for the parts that matter to ensure bindings take effect
- [FIX] Fixed incorrect parameter order in all scripts that use onSkill JS event

### DIFF
--- a/data/js/skill/anatomy.js
+++ b/data/js/skill/anatomy.js
@@ -6,7 +6,7 @@ function SkillRegistration()
 }
 
 /** @type { ( skillUse: BaseObject, skillUsed: number, objType: 0 | 1 ) => boolean } */
-function onSkill( pUser, objType, skillUsed )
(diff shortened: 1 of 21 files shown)


DragonSlayer62 (approved):
This very nice feature for scripting

DragonSlayer62 (approved):

PR #447, opened 2025-11-07
Mannequin

DragonSlayer62:
- [ADD] Added placeable Mannequin NPCs via deed. Players can equip mannequins with items, swap clothing with them, customize race and gender, and add a custom description


source/cPlayerAction.cpp
```diff
@@ -663,9 +663,18 @@ bool CPIEquipItem::Handle( void )
...
 	{
-		Bounce( tSock, i );
-		tSock->SysMessage( 1186 ); // You can't put items on other players!
-		return true;
+		TAGMAPOBJECT tMann  = k->GetTag( "Mannequin" );
+		bool isMann         = ( tMann.m_IntValue != 0 );
+		TAGMAPOBJECT tOwner = k->GetTag( "MannequinOwner" );
+		SERIAL ownerSer     = INVALIDSERIAL;
+
+		bool isOwner = ( ownerSer != INVALIDSERIAL && ownerSer == ourChar->GetSerial() ) || ( k->GetOwnerObj() == ourChar );
```
Xoduz: ownerSer is always equal to INVALIDSERIAL (it's set once on the line above, but never changed after that), so isOwner will never be true.

data/js/server/network/helpers/openpaperdoll.js
```diff
@@ -0,0 +1,14 @@
+//usage: TriggerEvent( 18000, "OpenPaperdoll", socket, targChar )
+
+function OpenPaperdoll( socket, targChar )
+{
+    // Create packet to open a paperdoll
+    var pStream = new Packet;
+    pStream.ReserveSize( 66 );
+    pStream.WriteByte( 0, 0x88 );
+    pStream.WriteLong( 1, targChar.serial );
+    pStream.WriteString( 5, targChar.name + ( targChar.title ? " " + targChar.title : "" ), targChar.name.length + ( targChar.title ? targChar.title.length + 1 : 0 ));
```
Xoduz: This needs to respect uox.ini settings for what to show in paperdolls, whether it's for a player or an NPC. Basically, to be a useful helper function, it needs to replicate the functionality provided via the following function in cPlayerAction.cpp:

> void PaperDoll( CSocket *s, CChar *pdoll )

Note that this helper function also will only work for characters that are within client's update range, and may or may not require a .Refresh( socket ) to work properly (see wholist.js script).

Alternatively, if you need this just for the mannequin, include this in the mannequin script instead of a separate helper script. Custom packets can be sent in regular scripts, they don't need to be "network scripts" for that.

data/js/server/misc/helpers/profanitycheck.js
```diff
@@ -0,0 +1,148 @@
+// Baseline single-word profanity (no slurs here; keep repo-friendly).
+// Add/remove to suit your shard's policy.
+let bad_words = [
```
Xoduz: Yeah, let's not do this. Make this read from an optional file of banned words instead (one per line), that shard admins can create or provide, but let's not host that (or a hard-defined list of profanities) in the UOX3 repo.

data/js/jse_fileassociations.scp
```diff
@@ -388,6 +389,15 @@
...
 15007=server/house/houseAddonUse.js
+15008=server/house/mannequindeed.js
+
+//-------------------------------------------
+// Helper Scripts [18000-18500]
+//-------------------------------------------
+18000=server/network/helpers/openpaperdoll.js
+18001=server/network/helpers/modifycontextmenu.js
+18002=server/misc/helpers/profanitycheck.js
+}
```
Xoduz: There seems to be an extra } here than there's supposed to be.

data/dfndata/npc/malevendors.dfn
```diff
@@ -782,6 +782,23 @@ NPCWANDER=0
 GET=m_playervendor f_playervendor
 }
 
+[mannequin]
```
Xoduz: Why is a mannequin a vendor?

data/dfndata/items/deeds/misc_deeds.dfn
```diff
@@ -76,4 +76,14 @@ morex=208
 name=a solen hive entrance deed
 id=0x14F0
 morex=209
+}
+
+[mannequindeed]
+{
+NAME=Mannequin Deed
```
Xoduz, line 93: get=base_item

Also, why is it weightless?

Xoduz (changes requested):
Don't forget to add the proper typescript comment headers to the scripts.

data/js/server/network/helpers/openpaperdoll.js
DragonSlayer62: I'll just add it to the mannequin it's not needed for anything more

data/js/server/misc/helpers/profanitycheck.js
DragonSlayer62: its more keep people from setting things on the manny but ill just remove it keep it as my own .

data/dfndata/npc/malevendors.dfn
DragonSlayer62: because I just used the player vendor settings to create it so acts like its owned by a player didn't need to cut into more stuff that was unneeded,

data/js/server/network/helpers/openpaperdoll.js
DragonSlayer62: removed openpaperdoll  just added to the manny

data/js/jse_fileassociations.scp
DragonSlayer62: okay remove the profanity check did normal description check and move this this to manny since only used on him.

data/dfndata/npc/malevendors.dfn
DragonSlayer62: moved into malehuman.dfn

data/dfndata/items/deeds/misc_deeds.dfn
DragonSlayer62: fixed this and add origin to is for hs when came out

PR #457, opened 2025-11-11
List of Fixes

DragonSlayer62:
- [FIX] Fixed Definitions.d.ts had misspelled void and onSpeechInput was missing SpeechID.
- [UPD] Updated lots of js scripts to use socket with system message where needed instead of pUser.
- [FIX] Fixed House Christmas Tree Items with a base ornament and Removed unused house_item line on the tree addon.
- [UPD] Updated Dictonary with missing golem construct messages and updated the js script to match.
- [UPD] Updated InteriorDecorator.js  gump with correct uox3 standard naming.


data/js/coreDefinitions.json
```diff
@@ -9005,14 +9015,14 @@
     },
     {
       "name": "onProfileRequest",
```
strattond: Missed one

```suggestion
      "name": "onProfileUpdate",
```

strattond (changes requested):

Xoduz (approved):

PR #478, opened 2025-12-30
New Equipment

DragonSlayer62:
- [ADD] Added IDs for more weapons, armors and clothing to material/weapon-type helper functions (weapontypes.js, material_types.js, combat.cpp)
- [UPD] Updated all the material type ids to be lower case to match (material_types.js)


strattond (approved):
I don't know how to review this, but it looks reasonable

PR #495, opened 2026-02-02
Extended DoMovingEffect, Updated GetSpawnRegions, Added showspawnregions command

Xoduz:
- [ADD] Added new GM command 'showspawnregions (js/commands/targeting/showspawnregions.js) which will visualize all spawn regions within range of targeted location for a short duration
- [ADD] Added new GM command 'spawnregionadmin (js/commands/spawnregionadmin.js) which shows a filterable list of all spawn regions in the game. Clicking the name of the spawn region in the menu will open up the tweak menu for said spawn region. Note that all changes done to spawn regions are temporary in nature, and only last until the next server restart (where it again reads from spawn region DFN files)
- [UPD] Updated 'tweak command to support modifying (temporary) of Spawn Regions. Can be accessed via spawnregionadmin command
- [UPD] Extended JS Function DoMovingEffect to support another optional parameter, a flag to specify if effect is stationary (true) at source location or not (false, default). Updated syntax:
  - DoMovingEffect( srcObject, trgObject, effect, speed, loop, explode, [hue], [renderMode], [stationaryFX] )
  - DoMovingEffect( srcObject, targX, targY, targZ, effect, speed, loop, explode, [hue], [renderMode], [stationaryFX] )
  - DoMovingEffect( srcX, srcY, srcZ, targX, targY, targZ, effect, speed, loop, explode, [hue], [renderMode], [stationaryFX] )
- [UPD] JS Function GetSpawnRegion has been split into two, one that accepts a singular spawn region ID param, and one that accepts coordinates for a location and returns all spawn regions found there:
  - GetSpawnRegion( spawnRegionID ) // Returns reference to specificed spawn region
  - GetSpawnRegions( x, y, worldnumber, instanceID ) // Returns list of references to all spawn regions at specified area
- [UPD] IterateOverSpawnRegions JS function now supports an optional socket parameter, which becomes available as a parameter in OnIterateSpawnRegion event if included
- [FIX] Fixed an issue with Spawn Region JS Objects where the .y1 property would return the same value as .x2
- [FIX] Fixed an issue with js/npc/pets/petbonding.js that prevented successful script compilation


data/js/commands/targeting/showspawnregions.js
```diff
@@ -0,0 +1,124 @@
...
+	let curX = ( typeof( tempShowNearby ) == "undefined" ? socket.GetWord( 11 ) : pChar.x );
+	let curY = ( typeof( tempShowNearby ) == "undefined" ? socket.GetWord( 13 ) : pChar.y );
+	delete socket.tempShowNearby; // Clean up
+	let curWorldNum = pChar.worldnumber;
+	let curInstanceID = pChar.instanceID;
+
+	// Check for spawn regions every 5 tiles in a radius of 20 tiles
+	let radius = 20;
+	let step = 5;
+	let foundRegs = {}; // Store unique regions
```
strattond, line 39: 👍

data/js/commands/targeting/showspawnregions.js
```diff
@@ -0,0 +1,124 @@
...
+
+	if( spawnRegs.length === 0 )
+	{
+	    socket.SysMessage( "No spawn regions found within 20 tiles." );
+	    return;
+	}
+
+	for( let i = 0; i < spawnRegs.length; i++ )
+	{
+		if( spawnRegs[i] )
```
strattond: Are you finding that GetSpawnRegions has returned a null object?  Would it make more sense to push a defensive non-null check up higher so you don't push a null region into spawnRegs in the first place?

That way you avoid a potentially non-0 list of spawnRegs which may only be null

data/js/commands/targeting/showspawnregions.js
```diff
@@ -0,0 +1,124 @@
...
+			let spawnX2 = spawnReg.x2;
+			let spawnY2 = spawnReg.y2;
+			let spawnZ = 0;
+			socket.SysMessage( "SpawnRegion #" + spawnReg.regionNum + ": " + spawnRegName + " (" + spawnX1 + "," + spawnY1 + " -> " + spawnX2 + "," + spawnY2 + ")" );
+			for( let x = spawnX1; x <= spawnX2; x++ )
+			{
+				if( DistanceBetween( curX, curY, x, spawnY1 - 1 ) <= 30 )
+				{
+					spawnZ = GetMapElevation( x, spawnY1 - 1, curWorldNum );
+					DoMovingEffect( x, spawnY1 - 1, spawnZ, x, spawnY1 - 1, spawnZ, 0x7, 0, 0xFF, false, hue, 0x2, true );
```
strattond: Woof, 13 parms ... that's getting a bit chonky.

Wonder if there's a way to reduce it (e.g. a struct for position) such as { x: x, y: spawnY1 - 1, ... } and grab what is needed?

data/js/commands/targeting/tweak.js
```diff
@@ -466,6 +491,20 @@ function command_TWEAK( pSocket, cmdString )
 
 	// First check if a valid target serial has been set by code, for instance when calling
 	// tweak command via the hard-coded wholist gump
+
+	// First, handle tweaking of potential spawn region object
+	if( pSocket.xText2 )
```
strattond: Are only spawn region objects capable of an xtext2?

data/js/commands/targeting/tweak.js
```diff
@@ -3816,6 +3887,203 @@ function HandleAccountTarget( pSocket, myTarget )
 	accountGump.Free();
 }
 
+// Handle properties of spawn region targets
+function HandleSpawnRegionTarget( pSocket, myTarget )
```
strattond, line 3905: Type definitions here please!

```suggestion
/** @type { ( pSocket: Socket, myTarget: Character | Item | null ) => void } */
function HandleSpawnRegionTarget( pSocket, myTarget )
```

data/js/commands/spawnregionadmin.js
```diff
@@ -0,0 +1,578 @@
...
+var spawnRegionListForceUpdate = false;
+
+/** @type { ( socket: Socket, cmdString: string ) => void } */
+function command_SPAWNREGIONADMIN( socket, cmdString )
+{
+	socket.currentChar.SetTempTag( "spawnRegionAdminWorldFilter", -1 );
+	ShowSpawnRegionList( socket, cmdString, null, null );
+}
+
+function ShowSpawnRegionList( socket, cmdString, filteredList, filterString )
```
strattond: If we know it, type defs would be good here


```suggestion
/** @type { ( socket: Socket, cmdString: string, filteredList: any, filterString: string | null ) => void } */
function ShowSpawnRegionList( socket, cmdString, filteredList, filterString )
```

Having said that ... no call to ShowSpawnRegionList passes a filterList or filterString?

data/js/commands/spawnregionadmin.js
```diff
@@ -0,0 +1,578 @@
...
+			filterString = gumpData.getEdit( 0 ).replace(/^\s+/, '');
+			var filteredList = ApplyFilter( pUser, spawnRegionList, true, true, true, false, filterString );
+
+			DisplaySpawnRegionListGump( socket, filteredList, filterString, 0 );
+			break;
+		}
+		case 2: // Prev page
+		{
+			var pageNum = parseInt( pUser.GetTempTag( "spawnRegionAdminCurrentPage" ) );
+			if( !pageNum )
```
strattond, line 284: Should this be a isNaN check?

```suggestion
			if( isNaN( pageNum ) )
```

data/js/npc/pets/petbonding.js
```diff
@@ -25,7 +25,7 @@ function onDeathBlow( killedPet, petKiller )
 
 		if( ValidateObject( petOwner ) && petOwner.socket != null )
 		{
-			TriggerEvent( 3108, "SendNpcGhostMode", petOwner.socket, 0, killedPet.serial, 1  );}
+			TriggerEvent( 3108, "SendNpcGhostMode", petOwner.socket, 0, killedPet.serial, 1  );
```
strattond, line 28: 👍

strattond (changes requested):
Some good fixes in there (whoopsie doodle).

Couple of minor suggestions / questions in here

data/js/commands/targeting/showspawnregions.js
Xoduz: Fair point. Moved null check up!

data/js/commands/targeting/showspawnregions.js
Xoduz: Chonky is the word, yeah! 🫠 Will think about this on the side. Maybe split it up. DoMovingStaticEffect. HAHA

data/js/commands/targeting/tweak.js
Xoduz: Nope, that's a fair point. I've updated this to use some temp tags instead.

data/js/commands/targeting/tweak.js
Xoduz: Done

data/js/commands/spawnregionadmin.js
Xoduz: Yup, legacy of copy & pasting some stuff from another variant of this. Removed those parameters!

data/js/commands/spawnregionadmin.js
Xoduz: Ah, no, because GetTempTag returns 0 (!) if tag does not exist. I removed the parseInt though, which was pointless since an int is what is stuffed in there to begin with. Could also potentially check for 0 instead of a "falsy" value, but... meh. It works :D

PR #500, opened 2026-02-07
Event Manager, onScriptLoad, onStart, FindItemBySection

Xoduz:
- [ADD] Added new JS Event that runs once for all scripts after shard has fully started up:
  - onScriptLoad()
- [ADD] Added new JS Event that runs for all instances of scripts attached to objects after shard has fully started up:
  - onStart( myObj ) // myObj can be a Character, Item or TownRegion
- [ADD] Added new JS Function to find nearest item with a specific section ID at a given location:
  - FindItemBySection( x, y, z, world, sectionId, [instanceId] )
- [ADD] Added new Event Manager system that initializes on shard startup, which listens to event registrations from other scripts and schedules these events to be run once, a number of times with an interval, or recurring on specific day(s)/time of the week. See UOX3 documentation for full details. System can be enabled or disabled via a new setting in uox.ini:
  - EVENTMANAGERSYSTEM=0/1 // Defaults to enabled
- [ADD] Added a global JS constant to detect in scripts whether UOX3 runs in debug mode:
  - UOX_DEBUG_MODE // Ex: if( UOX_DEBUG_MODE )
- [FIX] Fixed an incorrectly named socket variable in js/item/potion.js


DragonSlayer62 (approved):

data/dictionaries/dictionary.CSY
```diff
@@ -3543,16 +3543,30 @@
 6321=Vypadá to, jako by byla čepel pokryta kyselou látkou.
 6322=Vypadá to, jako by na čepeli byla korozivní látka.
 6323=Vypadá to, jako by na čepeli byla slabě korozivní látka.
-// [6540-6550] Hudební skříňka Dawns
```
strattond, line 3547: Implication on this is that you're behind develop - looks like you need a sync up?

data/js/item/potion.js
```diff
@@ -135,7 +135,7 @@ function onUseChecked( pUser, iUsed )
 					if( iUsed.amount > 1 )
 					{
 						// Stack of potions
-						var explosionPotion = iUsed.Dupe( socket );
+						var explosionPotion = iUsed.Dupe( pSocket );
```
strattond, line 138: 👍

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+
+// Configuration
+const eventManagerEnabled = GetServerSetting( "EventManagerSystem" );
+const eventDataFolder	= "events";
+const eventDataFile		= "event_list.json";
+const eventManagerDef	= "eventmanagerobj";
+const eventManagerTimer	= 1000;
+
+// Global variables
+var gEventManagerObj	= null;
```
strattond, line 17: TypeScript / VS Code gets happier if you annotate this type

e.g.

```
/** @type { Item | Character | null } */
var gEventManagerObj	= null;
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+	}
+}
+
+// Check for Event Manager's heartbeat
+function CheckForHeartbeat()
+{
+	if( !ValidateObject( gEventManagerObj ))
+		return false;
+
+	return ( gEventManagerObj.GetJSTimer( 0, SCRIPT.script_id ) > 0 );
```
strattond, line 40: Not a problem for this PR, but we need to define `script_id` on SCRIPT.

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+// One-time Initialization of Event Manager
+function InitEventManager()
+{
+	let oldEventManager = FindItemBySection( 1000, 1000, -50, 0, eventManagerDef );
+	if( ValidateObject( oldEventManager ))
+	{
+		// We found an existing event manager object in the world, use it!
+		gEventManagerObj = oldEventManager;
+		gEventManagerSerial = oldEventManager.serial;
+		if( UOX_DEBUG_MODE )
```
strattond, line 66: We'll need to define this constant in typescript too

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+					socket.SysMessage( GetDictionaryEntry( 6513, socket.language )); // Invalid event index. To see full event list: 'em list
+				}
+			}
+			break;
+		}
+		case "remove":
+		{
+			if( param == null || isNaN( parseInt( param )))
+			{
+				socket.SysMessage( GetDictionary( 6521, socket.language )); // Usage: 'em remove <index from event list>
```
strattond: Bug - this should be `GetDictionaryEntry`

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+const eventManagerEnabled = GetServerSetting( "EventManagerSystem" );
+const eventDataFolder	= "events";
+const eventDataFile		= "event_list.json";
+const eventManagerDef	= "eventmanagerobj";
+const eventManagerTimer	= 1000;
+
+// Global variables
+var gEventManagerObj	= null;
+var gEventManagerSerial	= 0;
+var gEventList			= [];
```
strattond, line 21: Likewise:
```
/** @type { any[] } */
var gEventList			= [];
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+		eventManager.StartTimer( eventManagerTimer, 0, SCRIPT.script_id );
+	}
+	else
+	{
+		Console.Print( GetDictionaryEntry( 6506 ) + "\n" ); // Event Manager: All events finished. Entering standby mode.
+	}
+}
+
+// Trigger the event via callback function
+function ExecuteCallback( callbackScriptID, callbackFunc, argsStart )
```
strattond, line 233:
```
/** @type { ( callbackScriptID: number, callbackFunc: any, argsStart: any[]  ) => void } */
function ExecuteCallback( callbackScriptID, callbackFunc, argsStart )
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+		consoleMsg = consoleMsg.replace( /%i/gi, callbackScriptID );
+		Console.Warning( consoleMsg );
+		return;
+	}
+
+	TriggerEvent( callbackScriptID, callbackFunc, argsStart );
+}
+
+// Validate conditions for event to run
+function CheckConditions( event )
```
strattond, line 250:
```
/** @type { ( event: any  ) => boolean } */
function CheckConditions( event )
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+		// Today's not the day. Try again next day
+		return false;
+	}
+
+	// No specific conditions are needed for the event to run
+	return true;
+}
+
+// Calculate when event should run next (for ones that run multiple times)
+function CalculateNextRun( event )
```
strattond, line 276:
```
/** @type { ( event: any  ) => void } */
function CalculateNextRun( event )
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+			}
+		}
+
+		event.nextRun = nextDate.getTime();
+	}
+}
+
+// Commands triggered via commands/em.js
+// Example usage: 'em list
+function EventManagerCommands( socket, command )
```
strattond, line 315:
```
/** @type { ( socket: Socket, command: string ) => boolean } */
function EventManagerCommands( socket, command )
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+		return false;
+
+	if( !eventManagerEnabled )
+	{
+		socket.SysMessage( GetDictionaryEntry( 6508 )); // Event Manager is currently disabled. Enable it via uox.ini
+		return false;
+	}
+
+	let args = command.split( " " );
+	subCmd = args[0].toLowerCase();
```
strattond:
```
let subCmd = args[0].toLowerCase();
```
Not defined explicitly

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+		default:
+			socket.SysMessage( GetDictionaryEntry( 6515, socket.language )); // Supported Event Manager commands: list, enable <#>, disable <#>, remove <#>, reset
+			break;
+	}
+	return false;
+}
+
+// Register a new Event
+// Usage from other scripts: TriggerEvent( [ManagerID], "RegisterEvent", { callbackScriptID: script_id, callbackFuncStart: "MyFunc", type: "INTERVAL", ... }, { argsStart }, { argsEnd } )
+function RegisterEvent( config, argsStart, argsEnd )
```
strattond, line 454:
```
/** @type { ( config: any, argsStart: any, argsEnd: any  ) => null | string } */
function RegisterEvent( config, argsStart, argsEnd )
```

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+		minute: config.minute || 0,
+
+		created: new Date().getTime()
+	}
+
+	// Initial calculation for next run
+	if( newEvent.type == "ONCE" )
+	{
+		// Run once at a very specific date/time
+		newEvent.nextRun = config.date || ( new Date().getTime() + 1000 );
```
strattond, line 518: This isn't initialized earlier on - is it worth ensuring that it's at least defined (maybe under created) and set it to unknown or null?

data/js/server/event_manager.js
```diff
@@ -0,0 +1,591 @@
...
+
+	return newEvent.id;
+}
+
+// Load events from .json file
+function LoadEvents()
+{
+	gEventList = [];
+
+	var mFile = new UOXCFile();
```
strattond, line 545: And we need to define UOXCFile at some point too

source/cScript.cpp
```diff
@@ -340,6 +341,13 @@ cScript::cScript( std::string targFile, UI08 rT, UI16 scrID ) : isFiring( false
 	JS_DefineFunctions( targContext, targObject, my_functions );
 	JS_SetPrivate( targContext, targObject, this );
 
+	// Define some global constants that can be used in any scripts
+#if defined( UOX_DEBUG_MODE )
```
strattond, line 345: 👍

source/cScript.cpp
```diff
@@ -476,7 +484,7 @@ JSBool cScript::InvokeEvent( const char* name, uintN argc, jsval* argv, jsval* r
 {
 	JSMapping->pushActive( this );
 #if defined UOX_DEBUG_MODE
-	Console.Warning( oldstrutil::format( "Triggering event '%s' from script %d", name, GetScriptID() ) );
+	Console.Log( oldstrutil::format( "Triggering event '%s' from script %d", name, GetScriptID() ) );
```
strattond, line 487: 👍

source/cServerData.h
```diff
@@ -212,7 +212,7 @@ class CServerData
 
 	// Once over 62, bitsets are costly.  std::vector<bool> has a special exception in the c++ specificaiton, to minimize wasted space for bools
 	// These should be updated
-	std::bitset<120>	boolVals;			// Many values stored this way, rather than using bools.
+	std::bitset<121>	boolVals;			// Many values stored this way, rather than using bools.
```
strattond, line 215: ☣️ 
That's ... a lot

source/findfuncs.cpp
```diff
@@ -590,6 +590,43 @@ CItem *FindItemNearXYZ( SI16 x, SI16 y, SI08 z, UI08 worldNumber, UI16 id, UI16
 	return currItem;
 }
 
+//o------------------------------------------------------------------------------------------------o
+//|	Function	-	FindItemBySectionNearXYZ()
+//o------------------------------------------------------------------------------------------------o
+//|	Purpose		-	Find items near specified location
```
strattond: It only finds one item, not multiple

source/findfuncs.cpp
```diff
@@ -590,6 +590,43 @@ CItem *FindItemNearXYZ( SI16 x, SI16 y, SI08 z, UI08 worldNumber, UI16 id, UI16
...
+//o------------------------------------------------------------------------------------------------o
+CItem *FindItemBySectionNearXYZ( SI16 x, SI16 y, SI08 z, UI08 worldNumber, const std::string &sectionId, UI16 instanceId )
```
strattond: Should this be named something more like `FindNearestItemBySectionNearXYZ`?

source/findfuncs.cpp
```diff
@@ -590,6 +590,43 @@ CItem *FindItemNearXYZ( SI16 x, SI16 y, SI08 z, UI08 worldNumber, UI16 id, UI16
...
+		if( toCheck == nullptr ) // no valid region
+			continue;
+
+		auto regItems = toCheck->GetItemList();
+		for( const auto &itemCheck: regItems->collection() )
+		{
+			if( !ValidateObject( itemCheck ) || itemCheck->GetInstanceId() != instanceId )
+				continue;
+
+			if( itemCheck->GetSectionId() == sectionId && itemCheck->GetZ() == z )
```
strattond: This means it has to be exactly on the same height plane, yes?  So a nearer item that is off by 1 in height wouldn't be counted?

So it's a 2D nearest neighbour check, not a 3D one?

strattond (changes requested):
Mostly all good.  Probably needs a sync up, one obvious bug (a function that doesn't exist).  The rest is largely cosmetic stuff.

source/cServerData.h
Xoduz: More, MORE

source/findfuncs.cpp
Xoduz: ~~Correct. It has a very niche use-case. Similar to FindItemNearXYZ(). For this event system, it fits the use-case 🫠For everything else.... we could use one of the area commands, I suppose.~~ I see what you mean. The comment refers to "find items", plural.

data/js/server/event_manager.js
Xoduz: Hm not sure if there's a point to it, as config.date is only referenced in this one particular case, and if I've understood it correctly, if it's undefined at the time of being assigned to .nextRun, the || operator will see it as a "falsy" value and instead default to Date().getTime() + 1000.

source/findfuncs.cpp
Xoduz: Yeah... maybe it's more useful to not be so strict about the Z. And use difference.MagSquared3D() instead of .Mag(). Something like this, maybe?

FindItemNearXYZ:
```c++
if( itemCheck->GetId() == id && abs( itemCheck->GetZ() - z ) <= 5 )
{
	Point3_st difference	= itemCheck->GetLocation() - targLocation;
	currDist = static_cast<UI16>( difference.MagSquared3D() );
	if( currDist < oldDist )
	{
		oldDist = currDist;
		currItem = itemCheck;
	}
}
```

FindItemBySectionNearXYZ:
```c++
if( itemCheck->GetSectionId() == sectionId && abs( itemCheck->GetZ() - z ) <= 5 )
{
	Point3_st difference	= itemCheck->GetLocation() - targLocation;
	currDist = static_cast<UI16>( difference.MagSquared3D() );
	if( currDist < oldDist )
	{
		oldDist = currDist;
		currItem = itemCheck;
	}
}
```

source/findfuncs.cpp
Xoduz: Done.

source/findfuncs.cpp
Xoduz: Updated comment, renamed function.

data/dictionaries/dictionary.CSY
Xoduz: WinMerge mixup.

data/js/server/event_manager.js
```diff
@@ -0,0 +1,594 @@
...
+			{
+				let index = parseInt( param );
+				if( index >= 0 && index < gEventList.length )
+				{
+					let newState = ( subCmd == "enable" );
+					let curEvent = gEventList[index];
+					curEvent.enabled = newState;
+
+					// Reschedule event to next valid window if it missed previous cycle
+					if( newState === true && event.nextRun < new Date().getTime() )
```
strattond, line 365: Do you mean `curEvent` here?  `event` doesn't exist?

data/js/server/event_manager.js
```diff
@@ -0,0 +1,594 @@
...
+					{
+						if( !curEvent.isRunning )
+						{
+							socket.SysMessage( GetDictionaryEntry( 6520, socket.language )); // Event reenabled and rescheduled for next valid activation window.
+							CalculateNextRun( event );
```
strattond, line 370: Likewise here - `event` vs `curEvent`

data/js/server/event_manager.js
```diff
@@ -0,0 +1,594 @@
...
+			{
+				socket.SysMessage( GetDictionaryEntry( 6521, socket.language )); // Usage: 'em remove <index from event list>
+			}
+			else
+			{
+				let index = parseInt( param );
+				if( index >= 0 && index < gEventList.length )
+				{
+					let curEvent = gEventList[index];
+					if( event.isRunning )
```
strattond, line 410: And here - do you mean `curEvent`?

data/js/server/event_manager.js
```diff
@@ -0,0 +1,594 @@
...
+		mFile.Free();
+	}
+}
+
+// Save all events to a .json file
+function SaveEvents()
+{
+	let mFile = new UOXCFile();
+
+	if( mFile.Open( eventDataFile, "w", eventDataFolder ))
```
strattond: So, this isn't going to do what you want - have a look at `CFile_Open` in the C++ code.

It doesn't return a value at all, it's a void return.

strattond (changes requested):

data/js/server/event_manager.js
Xoduz: I've consolidated all of these as "event". Thanks!

data/js/server/event_manager.js
Xoduz: I've consolidated all of these as "event". Thanks!

data/js/server/event_manager.js
Xoduz: I've consolidated all of these as "event". Thanks!

data/js/server/event_manager.js
Xoduz: Ahh I knew that already, but this one slipped past because of a "helpful assistant" 👀 Good catch, thanks. Updated to checking state of mFile after the Open call instead.

PR #529, opened 2026-03-07
Fix for characters going missing from lists in houses and for lockdown issue beneath a floor with a door

Xoduz:
- [UPD] Bumped version number to 0.99.7c
- [FIX] Fixed issue where characters could go missing from co-owner lists, friends lists, guest lists or ban lists in player houses, if those characters were loaded after the house itself
- [FIX] Fixed an issue that prevented locking down of stacked items in player houses when those items were directly beneath a door on the next floor up

### DIFF
--- a/data/js/server/house/houseCommands.js
+++ b/data/js/server/house/houseCommands.js
@@ -1211,8 +1211,8 @@ function CheckForNearbyDoors( myTarget, itemToCheck, pSocket )
 	{
 		if(( itemToCheck.type == 12 || itemToCheck.type == 13 ))
 		{
-			//if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z > GetTileHeight( myTarget ))
-			if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z >= 20 )
+			//if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z >= 20 )
+			if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z > GetTileHeight( myTarget ))
 			{
 				// Ignore doors on floors above
 				return false;
--- a/source/cMultiObj.cpp
+++ b/source/cMultiObj.cpp
@@ -1233,61 +1233,30 @@ void CMultiObj::PostLoadProcessing( void )
 	// Ensure base class processing fires first
 	CItem::PostLoadProcessing();
 
-	// BAN list
-	for( UI32 ser : pendingBans )
+	// Helper lambda for resolving deferred Character/Item-lists
+	auto ResolveObjects = [this]( std::vector<SERIAL>& pendingList, auto lookupFunc, auto action ) 
 	{
-		CChar *bList = CalcCharObjFromSer( ser );
-		if( ValidateObject( bList ) ) AddToBanList( bList );
-	}
-	pendingBans.clear();
+		for( SERIAL ser : pendingList ) 
+		{
+			auto ptr = lookupFunc( ser );
+			if( ValidateObject( ptr )) 
+			{
+				( this->*action )( ptr );
+			}
+		}
+		pendingList.clear();
+	};
 
-	// OWNER list
-	for( UI32 ser : pendingOwners )
-	{
-		CChar *cList = CalcCharObjFromSer( ser );
-		if( ValidateObject( cList ) ) AddAsOwner( cList );
-	}
-	pendingOwners.clear();
+	// Process deferred character-based lists
+	ResolveObjects( pendingBans, CalcCharObjFromSer, &CMultiObj::AddToBanList );
+	ResolveObjects( pendingOwners, CalcCharObjFromSer, &CMultiObj::AddAsOwner );
+	ResolveObjects( pendingFriends, CalcCharObjFromSer, &CMultiObj::AddAsFriend );
+	ResolveObjects( pendingGuests, CalcCharObjFromSer, &CMultiObj::AddAsGuest );
+	ResolveObjects( pendingVendors, CalcCharObjFromSer, &CMultiObj::AddVendor );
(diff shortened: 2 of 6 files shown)


source/cMultiObj.h
```diff
@@ -7,6 +7,15 @@ class CMultiObj : public CItem
 protected:
 	std::string				deed;
 
+	// Temporary serial vectors for deferred loading
+	std::vector<UI32>		pendingBans;
```
strattond: I know they're the same, but it wouldn't hurt for these to actually be SERIAL rather than UI32 - I mean, the typedef is there for clarity.

strattond:
The logic for this all makes sense.  I don't expect there to be any problems, and you're clearing the vector after it's done.

I had one question/suggestion.  And another thought - Multi PostLoadProcessing occurs before Items and Characters - is this likely to be an issue?  I would expect not, but thought I'd ask.

data/js/server/house/houseCommands.js
```diff
@@ -1211,8 +1211,8 @@ function CheckForNearbyDoors( myTarget, itemToCheck, pSocket )
 	{
 		if(( itemToCheck.type == 12 || itemToCheck.type == 13 ))
 		{
-			//if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z > GetTileHeight( myTarget ))
-			if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z >= 20 )
+			//if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z >= 20 )
+			if( itemToCheck.z > myTarget.z && itemToCheck.z - myTarget.z > GetTileHeight( myTarget ))
```
strattond, line 1215: Just confirming - you're happy for the change from `>=` to `>` here?

source/cMultiObj.h
strattond: Might be worth changing UI32 to SERIAL here?  I know it's the same, but it's clarity of intent

source/cMultiObj.cpp
```diff
@@ -1247,6 +1223,73 @@ bool CMultiObj::HandleLine( std::string &UTag, std::string &data )
 	return rValue;
 }
 
+//o------------------------------------------------------------------------------------------------o
+//|	Function	-	CMultiObj::PostLoadProcessing()
+//o------------------------------------------------------------------------------------------------o
+//|	Purpose		-	Used to setup any pointers that may need adjustment following the loading of the world
+//o------------------------------------------------------------------------------------------------o
+void CMultiObj::PostLoadProcessing( void )
```
strattond, line 1231: Looks fine to me.  The coder in me sees a whole bunch of repetition and thinks there must be a cleaner way, but this certainly shows intent quite happily.

Xoduz:
> The logic for this all makes sense. I don't expect there to be any problems, and you're clearing the vector after it's done.
> 
> I had one question/suggestion. And another thought - Multi PostLoadProcessing occurs before Items and Characters - is this likely to be an issue? I would expect not, but thought I'd ask.

Looking at it I don't think there's an issue, although there does appear to be some "lockdown magic" happening in CBaseObject::PostLoadProcessing that attempted to "fix" broken links to multis that were loaded after the items, but the changes in this PR seems to make that code a bit redundant 🤔 Thoughts?

source/cMultiObj.h
Xoduz: Thanks, replaced for clarity

Xoduz: Thanks, replaced for clarity

data/js/server/house/houseCommands.js
Xoduz: Yeah I think this is fine. Don't want the locked down item's height to reach exactly to the level of the door above

source/cMultiObj.cpp
Xoduz: Fair point. Replaced with a helper lambda for less repetition and easier future maintenance :P

source/cMultiObj.cpp
```diff
@@ -1247,6 +1223,42 @@ bool CMultiObj::HandleLine( std::string &UTag, std::string &data )
...
+			if( ValidateObject( ptr )) 
+			{
+				( this->*action )( ptr );
+			}
+		}
+		pendingList.clear();
+	};
+
+	// Process deferred character-based lists
+	ResolveObjects( pendingBans, CalcCharObjFromSer, &CMultiObj::AddToBanList );
```
strattond, line 1251: That's the spirit!

strattond:
> > The logic for this all makes sense. I don't expect there to be any problems, and you're clearing the vector after it's done.
> > I had one question/suggestion. And another thought - Multi PostLoadProcessing occurs before Items and Characters - is this likely to be an issue? I would expect not, but thought I'd ask.
> 
> Looking at it I don't think there's an issue, although there does appear to be some "lockdown magic" happening in CBaseObject::PostLoadProcessing that attempted to "fix" broken links to multis that were loaded after the items, but the changes in this PR seems to make that code a bit redundant 🤔 Thoughts?

That's ... interesting.  Your fix is broadly more relevant, but that sort of implies the ownership should already have been resolved.  I'd be interested to know if removing all the code in CBO::PLP (except for the SpawnSerial and oldLoc*) and using your fix results in any change.  I don't *think* so,  It would trigger CItem::LockDown as well as flagging it as dirty, but that's  probably not a good anyway during a world load.

strattond (commented):
I'd be interested to see if removing a chunk of CBO::PLP would still have the right behaviour.

strattond (approved):
Xuri — 2:00 PM
The lead senior dev test junior QA person gave the thumbs up - items and secure containers still locked down after repeated server restarts 😛

DragonSlayer62:
HAHAHA

PR #585, opened 2026-08-18
Misc updates

Xoduz:
- Changed protoList from heap-allocated JS::RootedObjectVector *to a JS::PersistentRootedObjectVector* for rooting safety
- Removed scriptJSMap in favor of accessing scripts and scriptIDs directly via reserved slots
- Restored JS engine's ability to set max memory usage based on INI setting rather than being hardcoded to 32MB
- Fixed exceptions with OnDamage and OnCombatDamageCalc JS Events caused by mismatch of param initialization and assignment
- Replaced fetching script environment with getThis() with JSMapping->currentActive() in SE_RegisterSpell() and SE_RegisterKey() for more explicit state tracking
- Added ValidateObject() to socket JS property getter for characters to fix a server exception
- Disabled optimization for Debug builds in VS2022 project to allow debugging again


uogonemetal:
okay as i was posting this the server crashed,................ Just kidding seems tolook okay and didnt have any issues so far

PR #587, opened 2026-08-20
Bugfixes and enhancements

Xoduz:
- [FIX] Fixed object and socket validation issues that could lead to script/server crashes in various scripts: repeatingcmds.js, ethereal_warrior.js, exodus_clockwork.js, kappa.js, lord_oaks.js, plaguebeast.js, rikktor.js, savage.js, savage_shaman.js, semidar.js, yamandon.js, colossal_blow.js, life_drain_aura.js
- [UPD] NPCs no longer perform line-of-sight checks to offline player characters before dismissing them as invalid targets in IsValidAttackTarget() in ai.cpp
- [FIX] Fixed issue where gumptype item property was not always loaded correctly from worldfiles
- [UPD] In-game time is now tracked server side in 24h format, and the non-functional AMPM tracking has been removed
- [ADD] UOX3 now resends the Time packet (0x5B) on next server loop after 'settime GM command is used
- [ADD] Added 'storm GM command to temporarily override weather in current region and force a storm in next weather update
- [FIX] Fixed incorrect coordinates for [REGIONSPAWN 5728] (Yew Catacombs NE Basement - Undeads)
- [UPD] Revamped connectivity information in UOX3 console to be less spammy when clients connect and transition from login to game server, etc.


uogonemetal (approved):
Only thing I was wandering about with the removal of that ampm and any saved worlds will the conversion do anything to them, just wandering or will it just convert the time to the correct formatting. i didn't seem to have any issues but just wandering.

Xoduz:
> Only thing I was wandering about with the removal of that ampm and any saved worlds will the conversion do anything to them, just wandering or will it just convert the time to the correct formatting. i didn't seem to have any issues but just wandering.

Well. If your server's in-game time is between noon and midnight when you upgrade to new version that has those time-tracking changes, you will experience a one-time -12 hour offset. Server only counted hours to 12 and then reset to 0 while flipping ampm setting between 0 and 1 each time, but now just counts to 24 instead. You can prevent that "time skip" by ensuring your server is saved when in-game time is anywhere from midnight to noon. I didn't add in backwards compatibility for the ampm tag because it's a hassle for something that has a one-time workaround anyway 😛

PR #588, opened 2026-08-20
Added support for Compressed Gumps

Xoduz:
- [ADD] Added CPSendCompressedGumpMenu packet class for compressed gump (0xDD) packet, with support for multi-byte UTF-8 Unicode characters
- [ADD] Added SendVecsAsCompressedGump() helper for preparing and sending compressed gumps with fallback to uncompressed gumps for client versions older than v4.0.11c
- [UPD] Updated all hardcoded sending of gumps to send compressed/uncompressed gump based on client version
- [FIX] Fixed an issue with quest/tracking arrow packet (0x07) where tSock->ClientVersion() was incorrectly used to compare to CV_HS2D when it should be using tSock->ClientType()


uogonemetal (approved):
Far as I can tell my custom gumps all still worked with this pr. i wont know if this broke custom house pr until its merged and I update that pr. but again my custom gumps all was still working.

PR #592, opened 2026-08-23
Account Updates for Security and Bug fixes

uogonemetal:
- [ADD] Added optional PBKDF2-HMAC-SHA256 password hashing for UOX3 accounts
  - PASSWORDHASHINGENABLED=0 // Enables hashing for newly created and changed passwords
  - Uses a random 16-byte salt, 600,000 iterations and a 32-byte derived password digest
  - Existing plaintext passwords are automatically migrated on next server startup
  - Existing hashed passwords continue working if hashing is later disabled
- [ADD] Added configurable per-IP login throttling for invalid credentials and repeated automatic account creation
  - LOGINTHROTTLEENABLED=1 // Enables or disables login throttling
  - LOGINTHROTTLEMAXATTEMPTS=5 // Failed attempts allowed before throttling begins
  - LOGINTHROTTLEWINDOW=60 // Failed-attempt tracking window in seconds
  - LOGINTHROTTLEINITIALDELAY=2 // Initial throttle delay in seconds
  - LOGINTHROTTLEMULTIPLIER=2 // Multiplier applied after additional failures
  - LOGINTHROTTLEMAXDELAY=300 // Maximum throttle delay in seconds
  - LOGINTHROTTLEENTRYTTL=3600 // Inactive throttle-entry lifetime in seconds
  - Connections attempted during an active delay are immediately rejected
- [ADD] Added local console command (E) for resetting account passwords
  - Generates a random 12-character temporary password and saves the account immediately
  - The temporary password is displayed only in the local server console and is not saved to logs
  - Players can log in with the temporary password and change it through the Help gump
- [UPD] Added dedicated [accounts] section to UOX.INI so account-management and security settings are grouped together, and moved INTERNALACCOUNTCREATION and ACCOUNTFLUSH settings into this new section
- [UPD] Centralized account password creation and changes so console resets, automatic accounts, imported accounts, in-game account creation and Help-gump changes respect the password hashing setting
- [UPD] Added the new password hasher and login throttler files to the Visual Studio 2022 and CMake build configurations
- [UPD] Updated the UOX3 documentation with password migration, password recovery, login throttling, username restrictions and the new [accounts] settings
- [FIX] Restricted newly created account usernames to 4–30 letters, numbers, underscores or hyphens
  - Prevents path separators, .. traversal sequences and other invalid characters from escaping the account directory
- [FIX] Removed plaintext passwords from login, character-creation and account-creation logs and raw packet dumps
- [FIX] Corrected account-map validation so both the username and password must match the same account record
- [FIX] Corrected Help menu password change gump to enforce 17-character client limit and reject invalid input


data/js/server/misc/helpgump.js
```diff
@@ -359,35 +361,32 @@ function onGumpPress( pSock, pButton, gumpData )
 		case 20: // Change account password
 		{
 			var userAccount = pUser.account;
-			var password = gumpData.getEdit( 0 );
-			var confirmpassword = gumpData.getEdit( 1 );
+			var password = gumpData.getEdit( 0 ).trim();
+			var confirmPassword = gumpData.getEdit( 1 ).trim();
 			var letterNumber = /^[0-9a-zA-Z]+$/;
```
strattond: You could also include the range here, I believe

```suggestion
			var letterNumber = /^[0-9a-zA-Z]{5,17}$/;
```

strattond:
So as a general recommendation, I would suggest reading https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html for best practices

source/cAccountClass.cpp
```diff
@@ -760,10 +774,11 @@ void cAccountClass::WriteAccountSection( CAccountBlock_st& actbTemp, std::fstrea
 UI16 cAccountClass::AddAccount( std::string sUsername, std::string sPassword, const std::string &sContact, UI16 wAttributes )
 {
 	// First were going to make sure that the needed fields are sent in with at least data
-	if( sUsername.length() < 4 || sPassword.length() < 5 )
+	if( !IsSafeNewAccountName( sUsername ) || sPassword.length() < 5 )
```
strattond: You should be checking maximum length as well.
```suggestion
	if( !IsSafeNewAccountName( sUsername ) || sPassword.length() < 5 || sPassword.length<> > 17 )
```

source/cAccountClass.cpp
```diff
@@ -760,10 +774,11 @@ void cAccountClass::WriteAccountSection( CAccountBlock_st& actbTemp, std::fstrea
...
 	{
-		// Username and password must be 4 and 5 characters or more in length, respectively
-		Console.Log( oldstrutil::format( "ERROR: Unable to create account for username '%s' with password of '%s'. Username/Password to short", sUsername.c_str(), sPassword.c_str() ), "accounts.log" );
+		// Existing accounts are deliberately grandfathered when loaded. This policy
+		// applies only to new accounts so shard upgrades do not rename or move data.
+		Console.Log( oldstrutil::format( "ERROR: Unable to create account for username '%s'. Usernames must be 4-30 ASCII letters, digits, underscores, or hyphens, and passwords must be at least 5 characters.", sUsername.c_str() ), "accounts.log" );
```
strattond: "and no more than 17 characters long"

source/cAccountClass.cpp
```diff
@@ -922,6 +948,59 @@ UI16 cAccountClass::AddAccount( std::string sUsername, std::string sPassword, co
...
+		// A successful login is the safest time to migrate a legacy plaintext credential.
+		// Hashing failure must not lock an existing player out, so login still succeeds.
+		SetPassword( account, password );
+	}
+	return true;
+}
+
+bool cAccountClass::SetPassword( CAccountBlock_st& account, const std::string &password )
+{
+	if( account.wAccountIndex == AB_INVALID_ID || password.length() < 5 || password.length() > 30 )
```
strattond: This limit is inconsistent on password length enforcement.

```suggestion
	if( account.wAccountIndex == AB_INVALID_ID || password.length() < 5 || password.length() > 17 )
```

source/cAccountClass.cpp
```diff
@@ -1836,7 +1915,7 @@ CAccountBlock_st& cAccountClass::GetAccountByName( std::string sUsername )
 		{
 			CAccountBlock_st& actbId = J->second;
 			// Check to see that these both are equal where it counts.
-			if( actbId.sUsername == actbName.sUsername || actbId.sPassword == actbName.sPassword )
+			if( actbId.sUsername == actbName.sUsername && actbId.sPassword == actbName.sPassword )
```
strattond, line 1971: :octocat:

source/cConsole.cpp
```diff
@@ -954,6 +954,47 @@ auto CConsole::Poll() -> void
...
+		{
+			if( !value.empty() )
+			{
+				value.pop_back();
+				std::cout << "\b \b";
+				std::cout.flush();
+			}
+			continue;
+		}
+		if( std::isprint( key ) && value.length() < 30 )
```
strattond, line 989: Inconsistent password enforcement length

```suggestion
		if( std::isprint( key ) && value.length() < 17 )
```

source/cLoginThrottler.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "cLoginThrottler.h"
+#include "cSocket.h"
+
+#include <algorithm>
+
+cLoginThrottler LoginThrottler;
+
+UI32 cLoginThrottler::AddressKey( CSocket *socket ) const
+{
+	return ( static_cast<UI32>( socket->ClientIP1() ) << 24 )
```
strattond: Don't we already have a function for this somewhere>

source/cLoginThrottler.cpp
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}
+
+	if( delaySeconds > 0 )
+	{
+		Console.Log( oldstrutil::format( "Login throttle activated for [%i.%i.%i.%i] for %u seconds.", socket->ClientIP4(), socket->ClientIP3(), socket->ClientIP2(), socket->ClientIP1(), delaySeconds ), "accounts.log" );
+	}
+}
+
+void cLoginThrottler::Cleanup( const std::chrono::steady_clock::time_point &now, UI32 entryTtl )
```
strattond: Worth putting a note in here indicating that if anyone chooses to call this, it _must_ be mutex protected or we're in trouble.

source/CPacketReceive.cpp
```diff
@@ -189,10 +190,9 @@ void CPIFirstLogin::Log( std::ostream &outStream, bool fullHeader )
 		outStream << "[RECV]Packet   : CPIFirstLogin 0x80 --> Length: 62" << TimeStamp() << std::endl;
 	}
 	outStream << "UserId         : " << Name() << std::endl;
-	outStream << "Password       : " << Pass() << std::endl;
+	outStream << "Password       : [REDACTED]" << std::endl;
```
strattond, line 193: Given this is only for logging (which isn't the usual behaviour), is it worth them having access to understand if people are using particular passwords?

source/CPacketReceive.cpp
```diff
@@ -517,9 +532,8 @@ void CPISecondLogin::Log( std::ostream &outStream, bool fullHeader )
 
 	outStream << "Key Used       : " << Account() << std::endl;
 	outStream << "SID            : " << Name() << std::endl;
-	outStream << "Password       : " << Pass() << std::endl;
-	outStream << "  Raw dump     :" << std::endl;
-	CPInputBuffer::Log( outStream, false );
+	outStream << "Password       : [REDACTED]" << std::endl;
```
strattond, line 537: Similar - logging isn't common.

strattond (changes requested):
Overall the idea is good.  There's some inconsistencies in password length expectations / enforcement, and I really wouldn't recommend rolling your own code for the actual hashing portions.

source/cLoginThrottler.cpp
DragonSlayer62: not for accounts that icould find it was insta login and log out without delay

data/js/server/misc/helpgump.js
DragonSlayer62: I fixed this thanks for pointing it out

source/cAccountClass.cpp
DragonSlayer62: I fixed this thanks for pointing it out

source/cAccountClass.cpp
DragonSlayer62: I fixed this thanks for pointing it out

source/cLoginThrottler.cpp
DragonSlayer62: changed this you was right

source/cLoginThrottler.cpp
DragonSlayer62: changed this and added notes about it

data/js/server/misc/helpgump.js
strattond: It still looks the same?

source/cAccountClass.cpp
Xoduz, line 1398: @DragonSlayer62 Suggestion: Just auto-hash all passwords in one go immediately after loading all the accounts on startup. Also makes logins slightly faster after enabling hashing of passwords, since it doesn't need to be done on the fly during login. Example implementation:

```suggestion
	wImportCount = ImportAccounts();

	// If password hashing is enabled, migrate any plaintext accounts in memory
	if( cwmWorldState->ServerData()->PasswordHashingEnabled() )
	{
		Console.Log( "Checking accounts for plaintext password migration..." );
		UI16 migratedAccounts = 0;
		for( auto &entry : m_mapUsernameIdMap )
		{
			CAccountBlock_st &account = entry.second;
			if( !account.sPassword.empty() && !cPasswordHasher::IsPasswordHash( account.sPassword ))
			{
				std::string hashedPassword;
				if( cPasswordHasher::HashPassword( account.sPassword, hashedPassword ))
				{
					account.sPassword = std::move( hashedPassword );
					account.bChanged = true;
					++migratedAccounts;
				}
			}
		}
		Console.Log( oldstrutil::format( "Account password migration complete. %u plaintext account(s) migrated to PBKDF2 hashes.", migratedAccounts ));
		if( migratedAccounts > 0 )
		{
			Save(); // Write updated hashes to accounts.adm immediately
		}
	}
```

source/cServerData.cpp
```diff
@@ -5892,18 +5908,29 @@ auto CServerData::SaveIni( const std::string &filename ) -> bool
 		ofsOutput << "GARGOYLEMAXWEIGHTBONUS=" << GargoyleMaxWeightBonus() << '\n';
 		ofsOutput << "}" << '\n';
 
+		ofsOutput << '\n' << "[accounts]" << '\n' << "{" << '\n';
```
Xoduz: Please shift the [account] section to just above the [skill & stats] section.

source/cLoginThrottler.cpp
```diff
@@ -0,0 +1,122 @@
...
+		if( ++operationsSinceCleanup >= 256 )
+		{
+			CleanupLocked( now, entryTtl );
+			operationsSinceCleanup = 0;
+		}
+	}
+
+	if( delaySeconds > 0 )
+	{
+		Console.Log( oldstrutil::format( "Login throttle activated for [%i.%i.%i.%i] for %u seconds.", socket->ClientIP4(), socket->ClientIP3(), socket->ClientIP2(), socket->ClientIP1(), delaySeconds ), "accounts.log" );
```
Xoduz: It would be good if there also was an output to server console for this, so it becomes obvious that someone has been throttled. If account name can also be included in that message, even better, so we can see if the attempts are connected to specific accounts or just random brute-force attacks.

data/js/server/misc/helpgump.js
DragonSlayer62: when i tryed ingame it matches the core code

Xoduz (approved):

PR #594, opened 2026-08-25
Fixed staff characters colliding with offline player characters

Xoduz:
- [FIX] Offline player characters are now flagged to ignore mobile collisions, allowing staff to walk through them without stamina restrictions


uogonemetal (approved):
I never noticed this before good find

